PR #4, opened 2021-06-02
Fix unl updates

mayurbhandary:
Fixes unl updates where a validator is removed. 

### Context of Change
UNL changes that removed existing validators were not handled correctly. Validators that were removed remained flagged as part of the validator list. This PR fixes the issue by explicitly setting the unl property to null if the validator is no longer on the list. Tests were added to verify that this is functioning correctly.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


natenichols (approved):
LGTM 👍

natenichols:
IMO we should wait on the duplicate validator fix bug to be fixed, and then roll both into a `beta` before deploying.

What's your opinion?

mayurbhandary:
@natenichols, I agree.

PR #5, opened 2021-06-08
fix: removes duplicate validator scores

mayurbhandary:
Fixes duplicate scores and validator entries when manifests are changed. 

### Context of Change
When manifests change, the old validator entry stays in the database for 7 days before it is cleared out. This causes the `/network/validators/:publicKey/reports` endpoint to pick up the same daily score twice. I've included a check to make sure revoked signing_keys are not included in the daily_agreement query. I've also modified the saveManifest function to update the revoked column in the validators table as manifests are saved. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### Testing
Tested by using dummy values in local db. I deleted the manifest entries for Bithomp's validator and set the 'revoked' column on the old validator entry to false. Then I ran the service and made sure that 'revoked' changes to true for the revoked signing_key. I also checked to see that the

### DIFF
--- a/src/api/routes/v1/validator-report.ts
+++ b/src/api/routes/v1/validator-report.ts
@@ -49,10 +49,10 @@ function formatResponse(response: DatabaseResponse): ScoreResponse {
 }
 
 /**
- * Reads nodes from database.
+ * Gets all daily score reports for a validator
  *
- * @param master_key - Master_key of validator.
- * @returns Locations of nodes crawled in the last 10 minutes.
+ * @param master_key - Master key of validator.
+ * @returns A promise that resolves to an array of ScoreResponse 
  */
 async function getReports(master_key: string): Promise<ScoreResponse[]> {
   return query('daily_agreement')
@@ -68,6 +68,7 @@ async function getReports(master_key: string): Promise<ScoreResponse[]> {
       'validators.master_key',
     )
     .where('validators.master_key', '=', master_key)
+    .andWhere('validators.revoked','=','false')
     .then((resp: DatabaseResponse[]) => resp.map(formatResponse))
 }
 
@@ -84,7 +85,6 @@ export default async function handleValidatorReport(
 ): Promise<void> {
   try {
     const master_key = req.params.publicKey
-
     const scores: ScoreResponse[] = await getReports(master_key)
 
     const response = {
--- a/src/api/routes/v1/validator.ts
+++ b/src/api/routes/v1/validator.ts
@@ -96,8 +96,7 @@ async function getValidators(): Promise<ValidatorResponse[]> {
       'master_key',
       'revoked',
     ])
-    .whereNotNull('signing_key')
-    .orWhereNotNull('master_key')
+    .where('revoked','=','false')
     .orderBy(['master_key', 'signing_key'])
     .then((res: dbResponse[]) => res.map(formatResponse))
 }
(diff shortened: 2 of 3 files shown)


src/api/routes/v1/validator-report.ts
```diff
@@ -49,10 +49,10 @@ function formatResponse(response: DatabaseResponse): ScoreResponse {
 }
 
 /**
- * Reads nodes from database.
+ * Gets all daily score reports for a validator
  *
- * @param master_key - Master_key of validator.
- * @returns Locations of nodes crawled in the last 10 minutes.
+ * @param master_key - Master key of validator.
+ * @returns A promise that resolves to an array of ScoreResponse
```
natenichols, line 55: Did I miss this comment when initially making it?

src/shared/database/index.ts
```diff
@@ -270,22 +270,36 @@ async function handleRevocations(
   manifest: DatabaseManifest,
 ): Promise<DatabaseManifest> {
   // Mark all older manifests as revoked
-  query('manifests')
+  const revokedSigningKeys = await query('manifests')
```
natenichols, line 273: How often does HandleRevocations run?

src/api/routes/v1/validator-report.ts
natenichols: Or does it run on a new manifest?

natenichols:
Are we gonna try to unittest this?

natenichols:
Also, do we need to update the API endpoint for validators to ignore `.where("revoked")`.

I can't remember if it already does that

natenichols (changes requested):
Don't see any particular changes, but I left a couple of questions that I have.

src/api/routes/v1/validator-report.ts
mayurbhandary: I'm not sure the comment didn't reflect what the function was doing. It looked like it was copy pasted from the crawler functions. getReports here gets all of the daily score report.

src/shared/database/index.ts
mayurbhandary: It gets called whenever a manifest is saved.

src/api/routes/v1/validator.ts
```diff
@@ -96,8 +96,7 @@ async function getValidators(): Promise<ValidatorResponse[]> {
       'master_key',
       'revoked',
     ])
-    .whereNotNull('signing_key')
-    .orWhereNotNull('master_key')
+    .where('revoked','=','false')
```
mayurbhandary, line 99: @natenichols, see here re: checking for revoked validators. I believe we initially decided to display all validators even if they are revoked. But, since we only delete validators from the database after 7 days of inactivity, I thought it is best if we just don't display the revoked ones. On the livenet explorer, you can see that Bithomp is listed twice because the entry for the old (revoked) signing key is still in the validators table and won't get removed for 7 days. Alternatively, we could just modify the explorer to not display the revoked validators.

natenichols: Yup, that sounds good to me!

mayurbhandary:
@natenichols, yes we will need to unit test this. I was hoping to to that in a separate PR so that we can get this fix out first, but happy to write the tests now if you feel we should do that first.

natenichols (approved):
This looks good to me.

Want to create a ticket to roll out a v1.0.1 and redeploy? Or do you want to wait for a couple more bug fixes?

mayurbhandary:
Let's roll out a v1.0.1 now since the bug should be fixed.

PR #8, opened 2021-07-09
Health

mayurbhandary:
Adds connection count to health endpoint and adds PR template.

### DIFF
--- a/src/shared/database/index.ts
+++ b/src/shared/database/index.ts
@@ -72,6 +72,7 @@ async function setupCrawlsTable(): Promise<void> {
       table.string('ip')
       table.integer('port')
       table.string('ws_url')
+      table.boolean('connected')
       table.string('networks')
       table.string('type')
       table.integer('uptime')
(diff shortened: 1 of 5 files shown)


natenichols:
Any particular reason why you want the number of connections to persist? Why not just add a function that returns `connections.length` in `connections.ts`, and report that in your health endpoint?

natenichols:
Ohhh, is see, cuz you want to report it on the `./api` directory.

Yeah, I'm cool w/ persisting, will start a review

src/shared/database/index.ts
```diff
@@ -72,6 +72,7 @@ async function setupCrawlsTable(): Promise<void> {
       table.string('ip')
       table.integer('port')
       table.string('ws_url')
+      table.boolean('connected')
```
natenichols, line 75: This requires an alteration to the database. We'll want to be certain that we record the process for deploying to stage, and be careful deploying to prod.

natenichols: https://www.postgresql.org/docs/12/sql-altertable.html

natenichols (approved):
So, I'm approving, because this LGTM. 

Would you post an example request/response on your local branch to this thread, so we can see what this health check looks like.

mayurbhandary:
@natenichols, example response:
`{"count":"48"}`

PR #9, opened 2021-07-14
Fix deadlock

mayurbhandary:
Fixes deadlock error and health check reporting

### Context of Change
Postgres was giving a deadlock error because too many unsuccessful connection attempts were calling saveNodeWsUrl. This PR only calls saveNodeWsUrl for connections that were previously successful. It also fixes a reporting error on the health endpoint. Multiple nodes(different public keys) with the same ip address were inflating the connection count. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
The error can be reproduced by clearing the local db before running startConnectionsDev. This PR will resolve the error.


src/connection-manager/connections.ts
```diff
@@ -152,9 +151,10 @@ setInterval(() => {
 export default async function startConnections(): Promise<void> {
   if (!cmStarted) {
     cmStarted = true
+    await clearConnectionsDb();
```
natenichols, line 154: So, what happens when a websocket disconnects? Does disconnect still set `connected` in the DB to false?

src/connection-manager/connections.ts
```diff
@@ -64,16 +64,15 @@ async function setHandlers(ip: string, ws: WebSocket): Promise<void> {
...
       ws.terminate()
       resolve()
     })
     ws.on('error', () => {
       if (connections.get(ip)?.url === ws.url) {
         connections.delete(ip)
       }
-      void saveNodeWsUrl(ws.url, false)
       ws.terminate()
       resolve()
```
natenichols, line 77: Will an `error` cause the `close` callback handler to be called? Because if not, we should also mark the connection as disconnected on Error

natenichols (changes requested):
I have one quick question, then will give the 👍

src/connection-manager/connections.ts
mayurbhandary: Yes the error causes the close callback handler to be called.

natenichols (approved):
LGTM 👍

PR #10, opened 2021-07-15
Logger

mayurbhandary:
Uses Bunyan Logger to timestamp and assign levels to existing console logs. 

### Context of Change
See [Bunyan](https://www.npmjs.com/package/bunyan) for details on the logger. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release


natenichols (changes requested):
This all seems totally reasonable. Looked over all files and I'm very excited to have an actual logger running 😃 

Two things: 
- I would like to stage this for a couple days, to make sure the logger works as expected. Also, we'll need to update the `vhs-deploy` to set a log level
- Would you add the [no-console](https://eslint.org/docs/rules/no-console) rule to our linter config? Since we're using a real logger now, we can forbid using `console.log`...

natenichols (approved):
This looks so much better than just `console.logging` Well done!

PR #25, opened 2022-04-20
fix: fix issue where all nodes are in all networks

mvadari:

This PR adds some walls between the networks while crawling, for misconfigured or old nodes.

In order for a node to be considered a part of a network, it must satisfy the following conditions:
* `complete_ledgers` must be available and must be in the same range of ledgers (since mainnet, devnet, and testnet all have different current ledger indices)
* If available, the UNL must be for the same network

### Context of Change

In the current `crawls` db of the VHS, all nodes are marked with all 3 networks.

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/src/crawler/crawl.ts
+++ b/src/crawler/crawl.ts
@@ -14,7 +14,7 @@ const DEFAULT_PORT = 51235
 const IP_ADDRESS = /^::ffff:/u
 const BASE58_MAX_LENGTH = 50
 
-const LEDGER_RANGE = 10000
+const LEDGER_RANGE = 100000
 
 /**
  *
--- a/src/crawler/crawl.ts
+++ b/src/crawler/crawl.ts
@@ -97,7 +97,6 @@ class Crawler {
       network = 'dev'
       unl = config.vl_dev
     }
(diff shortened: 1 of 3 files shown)


src/crawler/crawl.ts
```diff
@@ -47,6 +50,51 @@ class Crawler {
...
+  ): boolean {
+    const newestLedger = Crawler.getRecentLedger(thisCompleteLedgers)
+    const nodeNewestLedger = Crawler.getRecentLedger(nodeCompleteLedgers)
+    if (newestLedger == null || nodeNewestLedger == null) {
+      return false
+    }
+
+    const intNewestLedger = parseInt(newestLedger, 10)
+    const intNodeNewestLedger = parseInt(nodeNewestLedger, 10)
+    if (
```
natenichols, line 72: IMO this is more readable. If I read this correctly this is just checking if we're in range.

```js
Math.abs(intNewestLedger - intNodeNewestLedger) < LEDGER_RANGE
```

natenichols (approved):
LGTM. One suggestion, but nothing to hold up approval.

PR #29, opened 2022-04-21
refactor: put networks in code instead of env vars

mvadari:

This PR moves the network info from env vars to a TypeScript object. This makes it way easier to work with the networks.

Changes:
* Deleted `ENTRIES`, `VL_MAIN`, `VL_TEST`, `VL_DEV` env vars, and added `MAINNET_P2P_ENTRY`
* Moved all the network data to a TypeScript object/module
* Handled all cascading effects (which simplified a lot of the code)

### Context of Change

It's hard to add networks or modify the set of networks when their data is stored in environment variables. It's all public information and doesn't need to be hidden.

### Type of Change

- [x] Refactor (non-breaking change that only restructures code)

## Test Plan

CI passes.


.github/workflows/node.js.yml
```diff
@@ -53,12 +53,7 @@ jobs:
...
-        ENTRIES: 's2.ripple.com,s.altnet.rippletest.net,s.devnet.rippletest.net'
         # Rippled Server with access to manifest command
         RIPPLED_RPC_ADMIN: 'https://xrpl.ws/'
-        # Validator lists for main, dev, and test networks
-        VL_MAIN: 'vl.ripple.com'
-        VL_TEST: 'vl.altnet.rippletest.net'
-        VL_DEV: 'vl.devnet.rippletest.net'
-  
\ No newline at end of file
+        # (Placeholder) Mainnet entry point for crawler (without https://).
```
JST5000, line 58: Why is this a placeholder? (What process updates it to a non-placeholder value?)

README.md
```diff
@@ -32,10 +32,8 @@ Alternatively, update your `.bashrc` or `.zshrc` to export the environment varia
 
 Here are some example values for some environment variables:
 
-- `ENTRIES`: your rippled node(s) [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) separated by a comma, for example `ENTRIES=s1.ripple.com,s2.ripple.com`
-- `VL_MAIN`: a mainnet validator domain, for example `VL_MAIN=vl.ripple.com`
-- `VL_TEST`: a testnet validator domain, for example `VL_TEST=vl.altnet.rippletest.net`
-- `VL_DEV`: a devnet validator domain, for example `VL_DEV=vl.devnet.rippletest.net`
+- `MAINNET_P2P_SERVER`: your mainnet peer-to-peer rippled node [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name)
+- `MAINNET_UNL`: a mainnet validator domain, for example `MAINNET_UNL=vl.ripple.com`
```
JST5000, line 36: I wasn't quite sure what `a mainnet validator domain` meant, but clicking the link it just seems to be a UNL. Maybe this would be clearer wording for folks with less context?

```suggestion
- `MAINNET_UNL`: a mainnet validator domain which hosts a UNL, for example `MAINNET_UNL=vl.ripple.com`
```

src/connection-manager/manifests.ts
```diff
@@ -162,41 +164,18 @@ export async function updateUnls(): Promise<void> {
...
-        await query('validators')
-          .whereIn('signing_key', keys)
-          .update({ unl: config.vl_dev })
-        await query('validators')
-          .whereNotIn('signing_key', keys)
-          .where('unl', '=', config.vl_dev)
-          .update({ unl: null })
-      }
+      const network = networks.filter((ntwk) => ntwk.network === name)[0]
+      const networkUNL = network.unls[0]
```
JST5000, line 171: Going from networks -> networkUNL is done twice like this, maybe it's worth making a helper function?

.github/workflows/node.js.yml
mvadari: It's a placeholder because something needs to be there so the tests will run properly, but it doesn't actually need to be anything valid.

src/shared/utils/networks.ts
```diff
@@ -0,0 +1,47 @@
...
+if (config.mainnet_unl == null) {
+  mainnetUnls = mainMainnetUnls
+} else {
+  mainnetUnls = [config.mainnet_unl]
+  mainMainnetUnls.forEach((unl) => {
+    if (unl !== config.mainnet_unl) {
+      mainnetUnls.push(unl)
+    }
+  })
+}
```
JST5000, line 22: I think this can be simplified using Sets.
```suggestion
const mainMainnetUnls = new Set(['vl.ripple.com', 'vl.xrplf.org', 'vl.coil.com'])
if(config.mainnet_unl != null) {
  mainMainnetUnls.add(config.mainnet_unl)
}
mainnetUnls = Array.from(mainMainnetUnls)
```

mvadari: That doesn't solve the problem of wanting `config.mainnet_unl` to be at the front of the array if it's not `null`

JST5000: Ah right :)

JST5000 (approved):
LGTM

natenichols (approved):
LGTM!

src/shared/utils/index.ts
```diff
@@ -90,19 +91,20 @@ function blobToValidators(blob: UNLBlob): Set<string> {
...
-    vl_main: blobToValidators(main_blob),
-    vl_test: blobToValidators(test_blob),
-    vl_dev: blobToValidators(dev_blob),
-  }
+export async function getLists(): Promise<Record<string, Set<string>>> {
+  const lists = {}
+  const promises: Array<Promise<void>> = []
+  networks.forEach(async (network) => {
+    promises.push(
+      fetchValidatorList(network.unls[0]).then((blob) => {
```
natenichols, line 99: Do we need to make any changes here, because the old values are `vl_main`, and the new network values are `main`?

mvadari: I just double-checked, but the only place where that's used was also refactored.

PR #31, opened 2022-04-29
fix: save master key on validation

mvadari:

Validations contain a master key. These should be logged when a validation is received. This PR fixes that. It also updates the architecture documentation.

### Context of Change

Master keys were only being logged when a manifest was received, is only sent when a signing key is changed.

### Type of Change

## Test Plan

CI passes.


khancode (approved):

JST5000 (approved):
LGTM

natenichols (approved):

PR #32, opened 2022-04-29
BREAKING: add new APIs to filter nodes and validators by network

mvadari:

This PR adds new API methods to filter nodes and validators by what network they are in (mainnet/testnet/devnet). The existing API calls are also modified a bit, to be named to better represent what they do.

### Context of Change

Prep for sidechain support

### Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Test Plan

CI passes.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -16,11 +16,13 @@ There are 3 folders in `src`, corresponding to the 3 processes that the VHS runs
   * `/network/validator_reports`: Returns scores for the nodes that it has crawled in the last day.
   * `/network/topology`: Returns information about all the nodes that the crawler has crawled in the last hour.
   * `/network/topology/nodes`: Same as above.
-  * `/network/topology/nodes/:publicKey`: Returns information about a specific node.
+  * `/network/topology/nodes/:network`: Same as above, filtered to only return the nodes that are a part of `network`.
+  * `/network/topology/node/:publicKey`: Returns information about a specific node.
   * `/network/validators`: Returns information about all the validators that the VHS is paying attention to.
-  * `/network/validators/:publicKey`: Returns information about a specific validator.
-  * `/network/validators/:publicKey/manifests`: Returns the manifests of a specific validator.
-  * `/network/validators/:publicKey/reports`: Returns more detailed information about the reliability of a specific validator.
+  * `/network/validators/:network`: Same as above, filtered to only return the validators that are a part of `network`.
+  * `/network/validator/:publicKey`: Returns information about a specific validator.
+  * `/network/validator/:publicKey/manifests`: Returns the manifests of a specific validator.
+  * `/network/validator/:publicKey/reports`: Returns more detailed information about the reliability of a specific validator.
 
 
 ## SQL Table Schemas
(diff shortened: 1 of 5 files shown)


JST5000:
Would it be possible/preferable to just alias the old api names to the new names for now and mark them as deprecated?

mvadari:
> Would it be possible/preferable to just alias the old api names to the new names for now and mark them as deprecated?

Done in #33 (this PR is based on top of that one). I've asked around and it doesn't seem like anyone else uses this API (plus migration is super easy if they do), so I'm thinking it might be okay to just do it.

ARCHITECTURE.md
```diff
@@ -16,11 +16,13 @@ There are 3 folders in `src`, corresponding to the 3 processes that the VHS runs
   * `/network/validator_reports`: Returns scores for the nodes that it has crawled in the last day.
   * `/network/topology`: Returns information about all the nodes that the crawler has crawled in the last hour.
   * `/network/topology/nodes`: Same as above.
-  * `/network/topology/nodes/:publicKey`: Returns information about a specific node.
+  * `/network/topology/nodes/:network`: Same as above, filtered to only return the nodes that are a part of `network`.
+  * `/network/topology/node/:publicKey`: Returns information about a specific node.
```
JST5000, line 20: For this, does the URL have the word "node" in it or a specific node's identifier? (How does someone specify which specific node should be returned?)

JST5000: Oh, is the publicKey the parameter?

src/api/routes/v1/nodes.ts
```diff
@@ -191,19 +191,26 @@ export async function handleNode(req: Request, res: Response): Promise<void> {
...
     if (Date.now() - cache.time > 60 * 1000) {
       await cacheNodes()
     }
 
+    const { network } = req.params
+    const nodes =
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Necessary here
+      network == null
+        ? cache.nodes
+        : cache.nodes.filter((node) => node.networks?.includes(network))
```
JST5000, line 208: Can a single node be following both testnet and mainnet? (What would be the case where the node has multiple networks?)

ARCHITECTURE.md
mvadari: The `:` before a word indicates that it's a variable/parameter

src/api/routes/v1/nodes.ts
mvadari: That's not possible - a node has to be specific to a network (that's what #27 fixes).

JST5000: If that's not possible, what does `node.networks` represent? (Why is it an array of network names instead of a single one?)

mvadari:
> I've asked around and it doesn't seem like anyone else uses this API (plus migration is super easy if they do), so I'm thinking it might be okay to just do it.

Update: only one person has said that they use this API, and is fine with making the migration (it _is_ a super simple migration to make)

src/api/routes/v1/nodes.ts
mvadari: It's not an array, it's a string - it's just searching the string. (I'm not quite confident enough in my fix to assume that it's rock solid)

JST5000: Ah that makes sense :)

JST5000 (approved):
LGTM

natenichols (approved):

PR #35, opened 2022-05-03
fix: support agreement scores for validators without master keys

mvadari:

This PR ensures that all validators that the VHS receives validations for are having their agreements tracked. This includes:
* validators that haven't changed their master key (and therefore haven't issued manifests)
* validators that don't have a master key (and just have a hardcoded signing key instead of using an ephemeral signing key)

NOTE: This will require a database migration, to rename the `master_key` fields in `daily_agreement` and `hourly_agreement` to `main_key`. I wouldn't technically call this a breaking change, though, since the database is internal.

### Context of Change

Previously, the VHS would only calculate agreement scores for validators that had manifests - which were the UNL validators and validators that have changed their signing keys (after a year and a half of the VHS in operation, most validators were having their agreement scores calculated).

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)=

## Test Plan

CI passes.


src/shared/database/index.ts
```diff
@@ -431,9 +432,10 @@ export async function getAgreementScores(
 export async function signingToMaster(
   signing_key: string,
 ): Promise<string | undefined> {
-  return query('manifests')
+  return query('validators')
     .select('master_key')
     .where({ signing_key })
+    .where('revoked', '=', 'false')
```
mvadari, line 438: This change means that validators that haven't updated their signing key yet, but have a master key, will be scored based on their master key. The `revoked = false` is an additional check so old revoked signing keys won't be under the same bucket.

test/chains/agreement.test.ts
```diff
@@ -39,27 +24,7 @@ describe('Agreement', () => {
     await query('manifests').delete('*')
   })
 
-  test('Does not calculate score without master_key', async () => {
```
mvadari, line 42: This test was removed because it wasn't really valid. The VHS was originally using `manifests` to determine master keys, but that's provided in validations as well, so it's easier to pull master keys from the validations themselves. This test originally checked that agreement scores wouldn't generate if there weren't values in `manifests`.

natenichols (approved):
The LGTM!

I'm not entirely sure how to go about updating the database for `master_key` to `main_key`. Try it in staging first.

src/shared/database/index.ts
```diff
@@ -447,13 +449,13 @@ export async function signingToMaster(
...
 async function getHourlyAgreementScores(
-  validator: string,
+  validator: ValidatorKeys,
   start: Date,
   end: Date,
 ): Promise<AgreementScore[]> {
   return query('hourly_agreement')
     .select(['agreement'])
-    .where({ master_key: validator })
+    .where({ main_key: validator.master_key ?? validator.signing_key })
```
JST5000, line 458: If you make `ValidatorKeys` a class you can add a property `ValidatorKeys.main_key` which does this calculation on each call to avoid repeating this check since you do it fairly often.

src/shared/database/index.ts
```diff
@@ -484,55 +486,89 @@ function calculateAgreementScore(scores: AgreementScore[]): AgreementScore {
...
-    .where({ master_key })
-    .update({ agreement_1hour: agreement })
-    .catch((err) => log.error('Error Updating 1 Hour Validator Agreement', err))
+  const { master_key, signing_key } = validator_keys
+  if (master_key) {
+    await query('validators')
+      .where({ master_key })
+      .update({ agreement_1hour: agreement })
+      .catch((err) =>
+        log.error('Error Updating 1 Hour Validator Agreement', err),
```
JST5000, line 505: Should these errors be slightly different so it's easy to tell whether it checked the `master_key` or `signing_key`? (Ex. `Error Updating 1 Hour Validator Agreement for master_key ${master_key}`

src/shared/database/index.ts
```diff
@@ -484,55 +486,89 @@ function calculateAgreementScore(scores: AgreementScore[]): AgreementScore {
...
+      .update({ agreement_24hour: agreement })
+      .catch((err) =>
+        log.error('Error updating 24 Hour Validator Agreement', err),
+      )
+  } else {
+    await query('validators')
+      .where({ signing_key })
+      .update({ agreement_24hour: agreement })
+      .catch((err) =>
+        log.error('Error updating 24 Hour Validator Agreement', err),
```
JST5000, line 540: ditto

JST5000 (approved):
Few comments, but logic looks good to me :)

src/shared/database/index.ts
mvadari: I think it's more readable this way, and it feels a bit overkill to have a class just for that in JS land

JST5000: Fair enough :)

PR #36, opened 2022-05-04
fix: remove https from UNL

mvadari:

Title says it all.

The UNL name pulled from `/crawl` has `https` appended to it. The UNL lists in the VHS do not.

### Context of Change

Crawls were finding 0 nodes.

### Type of Change

## Test Plan

Works locally. Modifies a test to check for this.


JST5000 (approved):
LGTM

PR #39, opened 2022-05-06
feat: use a table for networks instead of hardcoding

mvadari:

This PR switches from using a hardcoded list for the networks to using a database (which is initially populated by the hardcoded list). 

### Context of Change

Sidechain work - it no longer needs to be a code change to add another network

### Type of Change

## Test Plan

CI passes. Works locally.

## Future Tasks
* Add an API endpoint to add networks to the db

### DIFF
--- a/src/connection-manager/chains.ts
+++ b/src/connection-manager/chains.ts
@@ -1,6 +1,8 @@
-import { saveValidatorChains } from '../shared/database'
+import { QueryBuilder } from 'knex'
+
+import { query } from '../shared/database'
 import { Ledger, ValidationRaw, Chain } from '../shared/types'
-import { overlaps } from '../shared/utils'
+import { getLists, overlaps } from '../shared/utils'
 import logger from '../shared/utils/logger'
 
 const log = logger({ name: 'chains' })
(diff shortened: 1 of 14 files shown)


src/connection-manager/chains.ts
```diff
@@ -31,6 +33,39 @@ function addLedgerToChain(ledger: Ledger, chain: Chain): void {
...
+async function saveValidatorChains(chain: Chain): Promise<void> {
+  let id = chain.id
+  const lists = await getLists().catch((err) => {
+    log.error('Error getting validator lists', err)
+    return undefined
+  })
+  if (lists != null) {
+    Object.entries(lists).forEach(([network, set]) => {
+      if (overlaps(chain.validators, set)) {
+        id = network
```
JST5000, line 51: Is this guaranteed to only happen once? (Or can there be multiple matches? - In which case do we always want the last one on the list?)

mvadari: Rewriting this code is way out of the scope of this PR. I just had to move the function to avoid a circular dependency.

JST5000: Ah I see, I didn't realize it was just moved.

src/shared/utils/networks.ts
```diff
@@ -36,7 +36,7 @@ const networks: Network[] = [
   },
   {
     network: 'dev',
-    entry: 's.devnet.ripple.test.net',
+    entry: 's.devnet.rippletest.net',
     port: 51235,
     unls: ['vl.devnet.rippletest.net'],
   },
```
JST5000, line 42: Is this hard coded value for the networks still used anywhere if you've switched to the table lookup?

JST5000: (If it's just the database setup file, maybe it should move into there?)

ckniffen (approved):

src/shared/database/setup.ts
```diff
@@ -146,3 +152,26 @@ async function setupDailyAgreementTable(): Promise<void> {
...
+      table.string('unls')
+      table.primary(['entry'])
+    })
+    networks.forEach((network) => {
+      query('networks')
+        .insert({
+          entry: network.entry,
+          network: network.network,
+          port: network.port,
+          unls: network.unls.join(','),
```
JST5000, line 172: nit: These are in different orders

JST5000 (approved):
LGTM

PR #40, opened 2022-05-10
feat: add new API endpoint to fetch network ID

mvadari:

This PR creates a new API endpoint, `get_network`, that fetches the network ID of the network that a node is connected to. If it doesn't exist, it creates one, and immediately crawls the network.

The API endpoint takes 300-40ms to load for a rippled node that exists, and 6ish seconds for an endpoint that is not a rippled node (since it waits for the call to `/crawl` to error).

### Context of Change

This is necessary for sidechains - if the Explorer wants to visit a new sidechain that it hasn't been to before, it needs to add it to the VHS. And this must happen programatically.

### Type of Change

## Test Plan

CI passes. Tested several scenarios by hand.

### DIFF
--- a/src/api/routes/v1/get-network.ts
+++ b/src/api/routes/v1/get-network.ts
@@ -185,7 +185,7 @@ async function addNode(url: string, unl: string | null): Promise<string> {
 }
 
 /**
- * Add entry to new network.
+ * Fetch network ID from network entry. If the network doesn't exist, adds it with a new ID.
  *
  * @param req - Express request.
  * @param res - Express response.
--- a/src/api/routes/v1/get-network.ts
+++ b/src/api/routes/v1/get-network.ts
@@ -35,11 +35,11 @@ async function fetchCrawls(host: string): Promise<Crawl> {
  * @param unl - The UNL of the node.
  * @returns Whether the UNL of the node has been seen before.
  */
-async function isUnlRecorded(unl: string): Promise<boolean> {
+async function getNetworkFromUNL(unl: string): Promise<string | undefined> {
   const result = await query('networks')
     .select('network')
     .where('unls', 'like', `%${unl}%`)
-  return result.length > 0
+  return result.length > 0 ? result[0].network : undefined
 }
 
 /**
@@ -48,11 +48,13 @@ async function isUnlRecorded(unl: string): Promise<boolean> {
  * @param publicKey - The public key of the node.
  * @returns Whether the public key of the node has been seen before.
  */
-async function isPublicKeyRecorded(publicKey: string): Promise<boolean> {
+async function getNetworkFromPublicKey(
+  publicKey: string,
+): Promise<string | undefined> {
   const result = await query('crawls')
-    .select('public_key')
+    .select('public_key', 'networks')
     .where('public_key', '=', publicKey)
-  return result.length > 0
+  return result.length > 0 ? result[0].networks : undefined
 }
 
 /**
@@ -60,15 +62,17 @@ async function isPublicKeyRecorded(publicKey: string): Promise<boolean> {
  *
  * @param url - The URL endpoint of the node.
  * @param unl - The UNL of the node.
+ * @returns The ID of the new network.
  */
-async function addNode(url: string, unl: string | null): Promise<void> {
+async function addNode(url: string, unl: string | null): Promise<string> {
   const currentNetworks = await query('networks')
     .select('network')
     .orderBy('network')
   const maxNetwork =
     currentNetworks[currentNetworks.length - networks.length - 1]?.network ?? 0
+  const newNetwork = (Number(maxNetwork) + 1).toString()
   const network: Network = {
-    network: (Number(maxNetwork) + 1).toString(),
+    network: newNetwork,
     entry: url,
     port: 51235,
     unls: unl ? [unl] : [],
@@ -80,6 +84,8 @@ async function addNode(url: string, unl: string | null): Promise<void> {
 
   const crawler = new Crawler()
   void crawler.crawl(network)
+
+  return newNetwork
 }
 
 /**
@@ -101,26 +107,35 @@ export default async function getNetwork(
 
     // check UNL
     const { node_unl } = crawl
-    if (node_unl != null && (await isUnlRecorded(node_unl))) {
-      return res.send({
-        result: 'error',
-        message: 'node UNL part of an existing network',
-      })
+    if (node_unl != null) {
+      const unlNetwork = await getNetworkFromUNL(node_unl)
+      // eslint-disable-next-line max-depth -- Necessary here
+      if (unlNetwork != null) {
+        return res.send({
+          result: 'success',
+          network: unlNetwork,
+          created: false,
+        })
+      }
     }
 
     // check if node public key is already recorded
     const { public_key } = crawl.this_node
-    if (await isPublicKeyRecorded(public_key)) {
+    const publicKeyNetwork = await getNetworkFromPublicKey(public_key)
+    if (publicKeyNetwork != null) {
       return res.send({
-        result: 'error',
-        message: 'node public key part of an existing network',
+        result: 'success',
+        network: publicKeyNetwork,
+        created: false,
       })
     }
     // add node to networks list
-    await addNode(entryUrl, node_unl)
+    const newNetwork = await addNode(entryUrl, node_unl)
 
     return res.send({
       result: 'success',
+      network: newNetwork,
+      created: true,
     })
   } catch (err) {
     return res.send({ result: 'error', message: err.message })
--- a/src/api/routes/v1/get-network.ts
+++ b/src/api/routes/v1/get-network.ts
@@ -210,7 +210,6 @@ export default async function getNetwork(
         return res.send({
           result: 'success',
           network: unlNetwork,
-          created: false,
         })
       }
     }
@@ -222,7 +221,6 @@ export default async function getNetwork(
       return res.send({
         result: 'success',
         network: publicKeyNetwork,
-        created: false,
       })
     }
     // add node to networks list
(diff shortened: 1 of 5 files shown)


src/api/routes/v1/get-network.ts
```diff
@@ -0,0 +1,202 @@
...
+ */
+async function addNode(url: string, unl: string | null): Promise<string> {
+  const currentNetworks = await query('networks')
+    .select('network')
+    .orderBy('network')
+  const maxNetwork =
+    currentNetworks[currentNetworks.length - networks.length - 1]?.network ?? 0
+  const newNetwork = (Number(maxNetwork) + 1).toString()
+  const network: Network = {
+    network: newNetwork,
```
JST5000, line 134: Nit: I got a bit confused reading this because I couldn't tell if `Network.network` was supposed to be an integer or a string. Maybe a better name for that field would be `id`?

src/api/routes/v1/get-network.ts
```diff
@@ -0,0 +1,202 @@
...
+}
+
+/**
+ * Add entry to new network.
+ *
+ * @param req - Express request.
+ * @param res - Express response.
+ * @returns The Express response.
+ */
+export default async function getNetwork(
```
JST5000, line 157: I wouldn't expect `getNetwork` to modify the network - Maybe `putNetwork` would be a better name?

JST5000: (Or maybe the description for this function should make it sound more like it's caching the result instead of adding a new network)

JST5000 (approved):
Logic looks good to me - Just a couple naming thoughts :)

src/api/routes/v1/get-network.ts
```diff
@@ -0,0 +1,202 @@
...
+        created: false,
+      })
+    }
+    // add node to networks list
+    const newNetwork = await addNode(entryUrl, node_unl)
+
+    return res.send({
+      result: 'success',
+      network: newNetwork,
+      created: true,
```
natenichols, line 197: I don't necessarily think we have to tell the user if an entry has been created.

natenichols (approved):

src/api/routes/v1/get-network.ts
mvadari: It's referred to as `network` in a bunch of the existing tables, which is why I kept that as `id`. But happy to rename it in this one if you'd prefer.

mvadari: It's actually a string (it's `main`/`test/`dev/`nft-dev` for the existing networks, and I didn't want to mess with the existing tables too much bc migrations by hand are hard and scary). But I thought the best way to add a new network without any details about it would be to just assign it a number.

mvadari: `currentNetworks.length - networks.length - 1` handles that - non-number strings will be after the number strings when sorted alphabetically

mvadari: Hm, looks like I'll have to rethink the sorting anyways - I forgot that with strings, '2' comes after '11'

src/api/routes/v1/get-network.ts
JST5000: That's fair, only name I could think of is `fetchNetwork`, but that's not much better - just the description update is clear enough for me :)

src/api/routes/v1/get-network.ts
JST5000: I think it'd be worth adding a comment explaining that going to `currentNetworks.length - networks.length - 1` is to go to the end of the named networks (`networks.length`) and into the numbered networks.

ckniffen (approved):

src/api/routes/v1/get-network.ts
mvadari: What about something like `getNetworkOrAdd`?

IMO the method name isn't that important, the API endpoint matters more

src/crawler/index.ts
```diff
@@ -49,4 +49,6 @@ async function main(): Promise<void> {
   }, LOCATE_INTERVAL)
 }
 
-main().catch((err: Error) => log.error(`${err.message}`))
+if (typeof require !== 'undefined' && require.main === module) {
```
ckniffen, line 52: Why is this required now?

ckniffen (approved):

src/api/routes/v1/get-network.ts
JST5000: I think `getNetworkOrAdd` conveys the functionality more clearly

src/api/routes/v1/get-network.ts
```diff
@@ -0,0 +1,237 @@
...
+import { Crawl } from '../../../shared/types'
+import logger from '../../../shared/utils/logger'
+
+const log = logger({ name: 'api-get-network' })
+
+const CRAWL_PORTS = [51235, 2459, 30001]
+
+let maxNetwork: number
+
+async function getMaxNetwork(): Promise<void> {
```
JST5000, line 16: `getMaxNetwork` implies that it will return maxNetwork, I think this name better reflects it's behavior.
```suggestion
async function updateMaxNetwork(): Promise<void> {
```

src/api/routes/v1/get-network.ts
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Adds the node (and its corresponding network) to the networks table.
+ *
+ * @param url - The URL endpoint of the node.
+ * @param unl - The UNL of the node.
+ * @returns The ID of the new network.
+ */
+async function addNode(url: string, unl: string | null): Promise<string> {
+  const newNetwork = (Number(maxNetwork) + 1).toString()
```
JST5000, line 167: Why do you have to put `Number(maxNetwork)` instead of `maxNetwork` on its own?

src/api/routes/v1/get-network.ts
```diff
@@ -0,0 +1,237 @@
...
+    const { public_key } = crawl.this_node
+    const publicKeyNetwork = await getNetworkFromPublicKey(public_key)
+    if (publicKeyNetwork != null) {
+      return res.send({
+        result: 'success',
+        network: publicKeyNetwork,
+      })
+    }
+    // add node to networks list
+    const newNetwork = await addNode(entryUrl, node_unl)
```
natenichols, line 227: Can we add a check that the crawl was valid (i.e. produced > `n`) nodes. Otherwise I think I could just spam random entries that will be saved in `addNode`. 

Also, I think `node_unl` can be undefined here (if node_unl == null). Should that be a possibility, or can we save an IP w/o a node. 

We should ensure that we got a valid crawl, a valid UNL, and a valid public_key before we save a node, else anyone can write arbitrary information to the database.

src/api/routes/v1/get-network.ts
```diff
@@ -0,0 +1,237 @@
...
+      if (unlNetwork != null) {
+        return res.send({
+          result: 'success',
+          network: unlNetwork,
+        })
+      }
+    }
+
+    // check if node public key is already recorded
+    const { public_key } = crawl.this_node
```
natenichols, line 218: is `this_node` guaranteed to exist? If not this will error:
```js
const foo = {}
const { does_not_exist } = foo.not_a_field
```
Because that the same as
```js
const does_not_exists = foo.not_a_field.does_not_exist
``` 

and JS doesn't let you call `.does_not_exist` on undefined.

src/api/routes/v1/get-network.ts
natenichols: I would also err on the side of making it more difficult to add your chain to the VHS database, over making it too easy.

src/api/routes/v1/get-network.ts
```diff
@@ -0,0 +1,237 @@
...
+    port: 51235,
+    unls: unl ? [unl] : [],
+  }
+  await query('networks').insert({
+    ...network,
+    unls: network.unls.join(','),
+  })
+
+  const crawler = new Crawler()
+  void crawler.crawl(network)
```
natenichols, line 182: So we kick off a crawl for that network to find new nodes for the newly added Node?

natenichols (changes requested):
This looks great! I have a couple questions about `getNetwork(` specifically, since that seems the core functionality that gets added.

Happy to approve once questions are answered.

src/crawler/index.ts
mvadari: It was running when I imported from this file. This prevents that from happening (it's the JS equivalent of `if __name__ == '__main__'` in python).

src/api/routes/v1/get-network.ts
```diff
@@ -0,0 +1,237 @@
...
+      })
+    }
+    // add node to networks list
+    const newNetwork = await addNode(entryUrl, node_unl)
+
+    return res.send({
+      result: 'success',
+      network: newNetwork,
+    })
+  } catch (err) {
```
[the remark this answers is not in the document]
mvadari: The app doesn't crash, but it does return an error if it's not caught.

src/api/routes/v1/get-network.ts
mvadari: Yes. Otherwise, you'll have to wait until the next set of crawls before you have the nodes from the network.

src/api/routes/v1/get-network.ts
mvadari: I don't quite understand how you could spam random entries - the `getNetworkFromPublicKey` call checks whether the node already exists in the known networks.

src/api/routes/v1/get-network.ts
mvadari: If it's a valid crawl, it should. If not, I believe it'll error in the [`crawlNode` function in the crawler](https://github.com/ripple/validator-history-service/blob/95ba58da44d82308e34c169e79cdad109528ab50/src/crawler/network.ts#L41-L95), where the crawl is fetched.

natenichols (approved):

PR #46, opened 2022-08-08
Add release version to validators

pdp2121:

Track validator server version by analyzing validation. 

### Context of Change

rippled version 1.8+ includes rippled version numbers in validation messages. Track these versions in the VHS API.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Validators relation did not have server version before. Changes were made to include this new field.

Example API call after the change:
`/network/validator/nHUvzia57LRXr9zqnYpyFUFeKvis2tqn4DkXBVGSppt5M4nNq43C`

Example result:
<code>{
    "validation_public_key": "nHUvzia57LRXr9zqnYpyFUFeKvis2tqn4DkXBVGSppt5M4nNq43C",
    "signing_key": "n9KY4JZY11ndNbg55dThnoQdU9dii5q3egzoESVXw4Z7hu3maCba",
    "master_key": "nHUvzia57LRXr9zqnYpyFUFeKvis2tqn4DkXBVGSppt5M4nNq43C",
    "domain": "digifin.uk",
    "chain": "main",
    <b>"server_version": "1.9.2"</b>,
    "current_index": 73739542,
    "agreement_1h": {
        "missed": 1,
        "total": 930,
        "score": "0.99892",
        "incomplete": true
    },
    "agreement_24h": {
        "missed": 1,
        "total": 930,
        "score": "0.99892",
        "incomplete": true
    },
    "agreement_30day": {
        "missed": 1,
        "total": 930,
        "score": "0.99892",
        "incomplete": true
    },
    "partial": false,
    "unl": "vl.ripple.com",
    "revoked": false,
    "result": "success"
}</code>

## Test Plan

- Server version is inserted for validators table.
- Correct version is updated for the server_version column.
- Correct release (basic, rc, beta, etc.)
- server_version column will not be updated if there's no server version in the ledger.

### DIFF
--- a/src/api/routes/v1/validator.ts
+++ b/src/api/routes/v1/validator.ts
@@ -19,7 +19,7 @@ interface ValidatorResponse {
   domain: string
   chain: string
   current_index: number
-  server_version: string | null
+  server_version?: string
   agreement_1h: {
     missed: number
     total: number
@@ -69,7 +69,7 @@ interface dbResponse {
   current_index: string
   domain: string
   chain: string
-  server_version: string | null
+  server_version?: string
   master_key?: string
   signing_key: string
   revoked?: boolean
--- a/src/api/routes/v1/validator.ts
+++ b/src/api/routes/v1/validator.ts
@@ -19,6 +19,7 @@ interface ValidatorResponse {
   domain: string
   chain: string
   current_index: number
+  server_version: string | null
   agreement_1h: {
     missed: number
     total: number
@@ -68,6 +69,7 @@ interface dbResponse {
   current_index: string
   domain: string
   chain: string
+  server_version: string | null
   master_key?: string
   signing_key: string
   revoked?: boolean
@@ -94,6 +96,7 @@ async function getValidators(): Promise<ValidatorResponse[]> {
       'current_index',
       'domain',
       'chain',
+      'server_version',
       'master_key',
       'signing_key',
       'master_key',
@@ -170,6 +173,7 @@ function formatResponse(resp: dbResponse): ValidatorResponse {
     master_key: resp.master_key,
     domain: resp.domain,
     chain: resp.chain,
+    server_version: resp.server_version,
     current_index: Number(resp.current_index),
     agreement_1h: hour1_score,
     agreement_24h: hour24_score,
@@ -199,6 +203,7 @@ async function findInDatabase(
       'current_index',
       'domain',
       'chain',
+      'server_version',
       'master_key',
       'signing_key',
       'master_key',
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -114,7 +114,7 @@ async function updateDailyAgreement(
  * @param ledger_index - Index of current ledger.
  * @returns Boolean.
  */
-function preceedFlagLedger(ledger_index: string): boolean {
+function isPreceedingFlagLedger(ledger_index: string): boolean {
   return parseInt(ledger_index, 10) % 256 === 255
 }
 
@@ -186,7 +186,7 @@ class Agreement {
       hashes.set(validation.ledger_hash, Date.now())
       this.validationsByPublicKey.set(signing_key, hashes)
       const server_version =
-        preceedFlagLedger(validation.ledger_index) &&
+        isPreceedingFlagLedger(validation.ledger_index) &&
         decodeServerVersion(validation.server_version)
       const validator: Validator = {
         master_key: validation.master_key,
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -1,6 +1,4 @@
-/* eslint-disable no-negated-condition -- Negation is used for release version decoding. */
-/* eslint-disable no-nested-ternary -- Nested ternary is used for release version decoding. */
-/* eslint-disable no-bitwise -- Bitwise operation is used for release version decoding. */
+/* eslint-disable no-negated-condition, no-nested-ternary, no-bitwise -- Manipulations used for server version decoding. */
 
 import {
   getAgreementScores,
@@ -117,13 +115,13 @@ function preceedFlagLedger(ledger_index: string): boolean {
 }
 
 /**
- * Classify release version type.
+ * Classify server version type.
  *
  * @param beta - Extracted 8-bit beta identifier.
  * @throws InvalidVersionException.
  * @returns String.
  */
-function extractReleaseVersionType(beta: number): string {
+function extractServerVersionType(beta: number): string {
   if (beta >> 6 === 0) {
     throw new Error('Unknown release type')
   }
@@ -148,7 +146,7 @@ function extractReleaseVersionType(beta: number): string {
  * @throws InvalidVersionException.
  * @returns String.
  */
-function decodeReleaseVersion(server_version?: string): string {
+function decodeServerVersion(server_version?: string): string {
   if (!server_version) {
     return ''
   }
@@ -176,7 +174,7 @@ function decodeReleaseVersion(server_version?: string): string {
   const patch = buf[4]
   const beta = buf[5]
 
-  const betaString = extractReleaseVersionType(beta)
+  const betaString = extractServerVersionType(beta)
 
   const version = `${major}.${minor}.${patch}${betaString}`
   return version
@@ -249,7 +247,9 @@ class Agreement {
     if (!hashes.has(validation.ledger_hash)) {
       hashes.set(validation.ledger_hash, Date.now())
       this.validationsByPublicKey.set(signing_key, hashes)
-      const release_version = decodeReleaseVersion(validation.server_version)
+      const server_version =
+        preceedFlagLedger(validation.ledger_index) &&
+        decodeServerVersion(validation.server_version)
       const validator = {
         master_key: validation.master_key,
         signing_key,
@@ -257,8 +257,7 @@ class Agreement {
         current_index: Number(validation.ledger_index),
         partial: !validation.full,
         last_ledger_time: new Date(),
-        ...(preceedFlagLedger(validation.ledger_index) &&
-          release_version && { version: release_version }),
+        ...(server_version && { server_version }),
       }
 
       chains.updateLedgers(validation)
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -103,6 +103,82 @@ async function updateDailyAgreement(
 }
 
 /**
+ * Updates agreement for a validator.
+ *
+ * @param ledger_index - Index of current ledger.
+ * @returns Boolean.
+ */
+function preceedFlagLedger(ledger_index: string): boolean {
+  return parseInt(ledger_index, 10) % 256 === 255
+}
+
+/**
+ * Classify release version type.
+ *
+ * @param beta - Extracted 8-bit beta identifier.
+ * @throws InvalidVersionException.
+ * @returns String.
+ */
+function extractReleaseVersionType(beta: number): string {
+  if (beta >> 6 === 0) {
+    throw new Error('Unknown release type')
+  }
+
+  const isBeta = beta >> 6 === 1
+  const isRC = beta >> 6 === 2
+  const isRelease = beta >> 6 === 3
+
+  const prefix = isRelease ? '' : '-'
+  const releaseType = isBeta ? 'b' : isRC ? 'rc' : ''
+
+  const releaseNum = !isRelease ? beta & 0x3f : ''
+
+  const betaString = `${prefix}${releaseType}${releaseNum}`
+  return betaString
+}
+
+/**
+ * Decode server software version.
+ *
+ * @param server_version - Software server version in 64 bits integer.
+ * @throws InvalidVersionException.
+ * @returns String.
+ */
+function decodeReleaseVersion(server_version?: string): string {
+  if (!server_version) {
+    return ''
+  }
+
+  const num = BigInt(server_version)
+
+  const buf = Buffer.alloc(8)
+
+  buf.writeBigInt64BE(num)
+
+  if (buf.length !== 8) {
+    throw new Error('Size must be 8')
+  }
+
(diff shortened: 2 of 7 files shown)


src/connection-manager/agreement.ts
```diff
@@ -1,3 +1,7 @@
+/* eslint-disable no-negated-condition -- Negation is used for release version decoding. */
+/* eslint-disable no-nested-ternary -- Nested ternary is used for release version decoding. */
```
ckniffen, line 2: You can disable multiple rules on the same line by comma separating the rules.

src/connection-manager/agreement.ts
```diff
@@ -169,13 +249,16 @@ class Agreement {
     if (!hashes.has(validation.ledger_hash)) {
       hashes.set(validation.ledger_hash, Date.now())
       this.validationsByPublicKey.set(signing_key, hashes)
+      const release_version = decodeReleaseVersion(validation.server_version)
```
mvadari, line 252:
```suggestion
      const release_version = preceedFlagLedger(validation.ledger_index) && decodeReleaseVersion(validation.server_version)
```

src/connection-manager/agreement.ts
```diff
@@ -169,13 +249,16 @@ class Agreement {
...
       const validator = {
         master_key: validation.master_key,
         signing_key,
         ledger_hash: validation.ledger_hash,
         current_index: Number(validation.ledger_index),
         partial: !validation.full,
         last_ledger_time: new Date(),
+        ...(preceedFlagLedger(validation.ledger_index) &&
+          release_version && { version: release_version }),
```
mvadari, line 261:
```suggestion
        version: release_version,
```

src/shared/database/setup.ts
```diff
@@ -121,6 +121,7 @@ async function setupValidatorsTable(): Promise<void> {
       table.string('unl')
       table.string('domain')
       table.boolean('domain_verified')
+      table.string('version')
```
mvadari, line 124: Might be more specific to call this `rippled_version` or something

mvadari:
Please fill out the PR template.

src/connection-manager/agreement.ts
pdp2121: I did this so that version will not be updated if there's ledger with no validation. With the change version: undefined will replace an actual version. Any ideas?

pdp2121: Made some changes. Let me know what you think. Thanks Mayukha!

src/connection-manager/agreement.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-negated-condition, no-nested-ternary, no-bitwise -- Manipulations used for server version decoding. */
```
natenichols, line 1: Disable this for the specific lines, not for the file. Then reviewers can see exactly why this is being disabled.

Also, if google search for the error, you can get an example of allowed and disallowed code to fix linter errors

src/connection-manager/agreement.ts
```diff
@@ -102,6 +104,82 @@ async function updateDailyAgreement(
...
+}
+
+/**
+ * Decode server software version.
+ *
+ * @param server_version - Software server version in 64 bits integer.
+ * @throws InvalidVersionException.
+ * @returns String.
+ */
+function decodeServerVersion(server_version?: string): string {
```
natenichols, line 149: This should be pretty easy to test if you export this function, or put it in the `shared` or `utils` and import here. 

Either way, testing this function will help prevent any versions show up incorrectly.

src/connection-manager/agreement.ts
```diff
@@ -102,6 +104,82 @@ async function updateDailyAgreement(
...
+  const isBeta = beta >> 6 === 1
+  const isRC = beta >> 6 === 2
+  const isRelease = beta >> 6 === 3
+
+  const prefix = isRelease ? '' : '-'
+  const releaseType = isBeta ? 'b' : isRC ? 'rc' : ''
+
+  const releaseNum = !isRelease ? beta & 0x3f : ''
+
+  const betaString = `${prefix}${releaseType}${releaseNum}`
```
natenichols, line 138: nit: just 

```js
return `${prefix}${releaseType}${releaseNum}`
```

src/connection-manager/agreement.ts
```diff
@@ -102,6 +104,82 @@ async function updateDailyAgreement(
...
+  }
+
+  const num = BigInt(server_version)
+
+  const buf = Buffer.alloc(8)
+
+  buf.writeBigInt64BE(num)
+
+  if (buf.length !== 8) {
+    throw new Error('Size must be 8')
```
natenichols, line 161: I wouldn't throw, prefer to just return `null`

and then check for null as shown above.

src/shared/database/setup.ts
```diff
@@ -121,6 +121,7 @@ async function setupValidatorsTable(): Promise<void> {
       table.string('unl')
       table.string('domain')
       table.boolean('domain_verified')
+      table.string('server_version')
```
natenichols, line 124: We'll need to carefully deploy this to staging to see if this will correctly add this column. 

If not we'll have to manually go and add the column

src/connection-manager/agreement.ts
```diff
@@ -102,6 +104,82 @@ async function updateDailyAgreement(
...
+/**
+ * Decode server software version.
+ *
+ * @param server_version - Software server version in 64 bits integer.
+ * @throws InvalidVersionException.
+ * @returns String.
+ */
+function decodeServerVersion(server_version?: string): string {
+  if (!server_version) {
+    return ''
```
natenichols, line 151: I would probably prefer to return null, and handle the case where this version is null w/ 
```js
const decodedVersion = decodeServerVersion(...)
if (decodedVersion != null)
    await saveVersion(validator, decodedVersion)
```

src/connection-manager/agreement.ts
```diff
@@ -169,13 +247,17 @@ class Agreement {
...
+        preceedFlagLedger(validation.ledger_index) &&
+        decodeServerVersion(validation.server_version)
       const validator = {
         master_key: validation.master_key,
         signing_key,
         ledger_hash: validation.ledger_hash,
         current_index: Number(validation.ledger_index),
         partial: !validation.full,
         last_ledger_time: new Date(),
+        ...(server_version && { server_version }),
```
natenichols, line 260: This seems a little complicated... 

What type are we trying to do here. Should this be type `string | null`?

natenichols (changes requested):
Requesting fixes for to the uncaught throws and disabled linter rules.

src/connection-manager/agreement.ts
natenichols: If it's type string or null, just pass in `decodedVersion`

ckniffen (approved):

src/connection-manager/agreement.ts
```diff
@@ -102,6 +108,16 @@ async function updateDailyAgreement(
   })
 }
 
+/**
+ * Detect the ledger before a flagged ledger, which will contain server version information.
+ *
+ * @param ledger_index - Index of current ledger.
+ * @returns Boolean.
+ */
+function preceedFlagLedger(ledger_index: string): boolean {
```
natenichols, line 117: nit, would prefer if this was `preceedsFlagLedger()`

src/connection-manager/agreement.ts
```diff
@@ -178,6 +197,10 @@ class Agreement {
         last_ledger_time: new Date(),
       }
 
+      if (server_version) {
```
natenichols, line 200: nit: prefer `server_version != null`

src/shared/database/agreement.ts
```diff
@@ -225,3 +225,73 @@ export async function update30DayValidatorAgreement(
...
+  if (typeBits === 0) {
+    return null
+  }
+
+  const isBeta = typeBits === 1
+  const isRC = typeBits === 2
+  const isRelease = typeBits === 3
+
+  const prefix = isRelease ? '' : '-'
+  // eslint-disable-next-line no-nested-ternary -- Nested tenary is used to determine release type based on 2 conditions.
```
natenichols, line 248: Yea, I"m okay with this nested ternary.

natenichols (approved):
Looks great, nice work @pdp2121.

Left a couple of nits, but nothing that should hold up approval.

natenichols:
have we deployed this to staging yet?

src/connection-manager/agreement.ts
ckniffen: That is equal to what is there.  Did you instead mean `server_version !== null`?

natenichols: I prefer `!=`. Willing to hear other opinions, though.

`server_version != null` will not enter the `if ()` statement if we ever hit a case where `server_version` is `undefined`. 

On the off chance we hit an `undefined` server_version, we don't want to enter that if statement.

The following script:
```js
const foo = undefined;

if (foo !== null)
  console.log(`hit case !== with ${foo}`)

if (foo != null)
  console.log(`hit case != null with ${foo}`)
```

Outputs: 
```
hit case !== with undefined
```

Worth noting, I don't think this actually will change anything with the runtime behavior of this function. I _think_ `if (...)` evaluates a very similar way.

natenichols: Also worth noting that I no longer own this repo, so if there's a style guide y'all use you're free to disregard my opinions :)

ckniffen: Correct.  Most linting configs frown on `!=`.  I have not looked into this repos rules yet.

natenichols:
Also, is the plan to add the `verison` field to the API in another PR?

src/connection-manager/agreement.ts
mvadari: This should involve some verb like `is` or `does` - maybe `isPreceedingFlagLedger`? To make it clear that it's returning a boolean

src/connection-manager/agreement.ts
mvadari: I also prefer `server_version != null` - specifically with `null` you want to use just `==` or `!=` (I believe most linters respect that, we definitely have a bunch of those in xrpl.js). That way, it also catches `undefined` and similar null-like values.

src/shared/database/agreement.ts
```diff
@@ -225,3 +225,73 @@ export async function update30DayValidatorAgreement(
       )
   }
 }
+
+/**
+ * Classify server version type.
+ *
+ * @param beta - Extracted 8-bit beta identifier.
+ * @throws InvalidVersionException.
+ * @returns Release version type (basic, beta, rc, etc.).
```
mvadari, line 234: The paranthetical here should be in line 230 as well, IMO - I was a little confused as to what this method was doing at first, and the documentation should be clear about that.

mvadari: It actually looks like this documentation is incorrect - this implies that it's only returning the type, but it's also returning the version number.

src/shared/database/setup.ts
mvadari: This won't add the column, unless the table doesn't exist (obviously it already does in staging and prod). This is how to modify a column if the table already exists: https://github.com/ripple/validator-history-service/blob/main/src/shared/database/setup.ts#L142-#L145 (you could probably figure out how to modify this code to get it to add a column instead)

mvadari (changes requested):
a few little changes (nothing major), but they should really be addressed before merging

src/connection-manager/agreement.ts
pdp2121: Since `server_version = isPreceedingFlagLedger(validation.ledger_index) && decodeServerVersion(validation.server_version)`, it can also take a value of `false` when `isPreceedingFlagLedger(validation.ledger_index)` return `false`. So should it be `server_version != null && server_version !== false`?

src/shared/database/agreement.ts
pdp2121: It returns rc/beta number (1-63), e.g rc7, not the full version number. Should I rename the variable `releaseNum` and update the doc to avoid confusion?

src/shared/database/setup.ts
pdp2121: I added the column in the if statement on line 131.

src/connection-manager/agreement.ts
natenichols: If that's the case, I'm okay with `if (server_version)`

src/shared/database/agreement.ts
pdp2121: @natenichols

natenichols: I would say something like 

"Extracts the server version type from the beta version byte. This function will return nothing if a release version, `-b\<beta version\>` if a beta version, and `-rc<\release candidate version\>` if a release candidate."

instead of 

"Classify server version type"

natenichols: I actually prefer `extractServerVersionType`, because this helper does not extract the server_version, it only extracts the release type and version (i.e. nothing if release, -b1 if beta, -rc1 if release candidate).

But that's just my opinion, and am more than willing to discuss better names.

natenichols: maybe `extractReleaseType`?

src/api/routes/v1/validator.ts
```diff
@@ -68,6 +69,7 @@ interface dbResponse {
   current_index: string
   domain: string
   chain: string
+  server_version: string | null
```
natenichols, line 72: this can be 
```ts
server_version?: string
```

src/api/routes/v1/validator.ts
```diff
@@ -19,6 +19,7 @@ interface ValidatorResponse {
   domain: string
   chain: string
   current_index: number
+  server_version: string | null
```
natenichols, line 22: same

natenichols (approved):

mvadari (approved):

PR #37, opened 2022-05-05
feat: add NFT-Devnet to the list of networks

mvadari:
## High Level Overview of Change

Title says it all.

This PR can also serve as a template for how to add a new network to the VHS.

### Context of Change

This will allow us to support the networks page on the Explorer for NFT-Devnet.

### Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Test Plan

Works locally.

PR #51, opened 2022-10-04
Add amm network

pdp2121:

Add amm network to VHS.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/api/routes/v1/validator.ts
+++ b/src/api/routes/v1/validator.ts
@@ -268,11 +268,11 @@ async function getChains(
   if (unl == null) {
     return undefined
   }
-  const results = await query('networks')
-    .select('id')
+  const results = await query('validators')
+    .select('chain')
     .distinct()
-    .where('unls', '~*', `(%,)?${unl}(,%)?`)
-  return results.map((result) => result.id as string)
+    .where({ unl })
+  return results.map((result) => result.chain as string)
 }
 
 /**
--- a/src/api/routes/v1/validator.ts
+++ b/src/api/routes/v1/validator.ts
@@ -268,11 +268,11 @@ async function getChains(
   if (unl == null) {
     return undefined
   }
-  const results = await query('validators')
-    .select('chain')
+  const results = await query('networks')
+    .select('id')
     .distinct()
-    .where({ unl })
-  return results.map((result) => result.chain as string)
+    .where('unls', '~*', `(%,)?${unl}(,%)?`)
+  return results.map((result) => result.id as string)
 }
 
 /**
--- a/src/shared/database/setup.ts
+++ b/src/shared/database/setup.ts
@@ -190,4 +190,21 @@ async function setupNetworksTable(): Promise<void> {
         .catch((err: Error) => log.error(err.message))
     })
   }
+
+  const networksCount = await query('networks').countDistinct('id')
+  const networksIds = await query('networks').pluck('id')
+  if (networksCount[0].count < networks.length) {
+    networks.forEach((network) => {
+      if (!networksIds.includes(network.id)) {
+        query('networks')
+          .insert({
+            id: network.id,
+            entry: network.entry,
+            port: network.port,
+            unls: network.unls.join(','),
+          })
+          .catch((err: Error) => log.error(err.message))
+      }
+    })
+  }
 }
--- a/src/shared/database/setup.ts
+++ b/src/shared/database/setup.ts
@@ -89,21 +89,7 @@ async function setupManifestTable(): Promise<void> {
 }
 
 async function setupLedgersTable(): Promise<void> {
-  const hasLedgers = await db().schema.hasTable('ledgers')
-  if (!hasLedgers) {
-    await db().schema.createTable('ledgers', (table) => {
-      table.string('ledger_hash').primary()
-      table.integer('ledger_index').index()
-      table.integer('full')
-      table.integer('main')
-      table.integer('altnet')
-      table.json('partial')
-      table.json('missing')
-      table.double('avg_load_fee')
-      table.dateTime('avg_sign_time')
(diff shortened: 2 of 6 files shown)


mvadari (commented):
To add the AMM network, you only need the changes to `network.ts`. What's the purpose of all the other changes?

See: https://github.com/ripple/validator-history-service/pull/37

src/shared/database/setup.ts
```diff
@@ -190,4 +190,21 @@ async function setupNetworksTable(): Promise<void> {
         .catch((err: Error) => log.error(err.message))
     })
   }
+
+  const networksCount = await query('networks').countDistinct('id')
```
pdp2121, line 194: @mvadari This one is to manually add a new network into the table if not already existed.

src/api/routes/v1/validator.ts
```diff
@@ -268,11 +268,11 @@ async function getChains(
...
   }
-  const results = await query('validators')
-    .select('chain')
+  const results = await query('networks')
+    .select('id')
     .distinct()
-    .where({ unl })
-  return results.map((result) => result.chain as string)
+    .where('unls', '~*', `(%,)?${unl}(,%)?`)
+  return results.map((result) => result.id as string)
```
pdp2121, line 275: @mvadari This one is the change to the way we retrieve chain from unl to return validators for /v2/network/validators/{unl}. Previously the way we did it was to match any unl with chain in validators table but a lot of validators has unl of null now (in fact, all validators with chain amm-dev has unl of null). So I changed the way we retrieve chain from unl using network table and regex to avoid this situation. I feel like this is a better way of doing it.

pdp2121:
> To add the AMM network, you only need the changes to `network.ts`. What's the purpose of all the other changes?

Add inline comments to explain each of them. Please let me know if you have questions.

src/shared/database/setup.ts
mvadari: That automatically happens by adding it to `networks`

mvadari: https://github.com/ripple/validator-history-service/blob/main/src/shared/database/setup.ts#L176-L186

mvadari: Notice that in #37 a new network was added without this being necessary

src/api/routes/v1/validator.ts
mvadari: All the validators in that chain have a UNL of `null` because they didn't have a UNL. I don't understand why we should return validators with a UNL of `null` - that doesn't make sense, when we want all the chains that have the given UNL

mvadari: We don't currently really have support for networks without a UNL

src/shared/database/setup.ts
pdp2121: There's an `if (!hasNetworks) {` condition there, so if the network table already existed, it will skip that part I believe

src/api/routes/v1/validator.ts
pdp2121: Currently, we still use validators with unl of null in prod:
https://data.xrpl.org/v1/network/validators/vl.ripple.com

pdp2121: I'm confused as well

mvadari: It's a little easier to tell here: https://data.xrpl.org/v1/network/validators/vl.altnet.rippletest.net
The chain has at least one validator with the given UNL, and then all validators are pulled from those chains

src/shared/database/setup.ts
mvadari: Hm, that's fair - in that case, we shouldn't be doing this twice when the `networks` table is created, those two sections of the code should be combined.

src/api/routes/v1/validator.ts
pdp2121: Oh I see. So the logic for /v2/network/validators/{unl} would be first find the chains that has at least one UNL that match the requested unl then pull all validators from those chains? If so, the old logic would be correct then, but /v2/network/validators/vlamm.devnet.rippletest.net will return an empty list since none of those validators has an UNL yet.

mvadari: Exactly, yeah

mvadari: One thing to note: you'll have to manually remove the data for that network from the production instance of the VHS first, since it was accidentally already added

src/shared/database/setup.ts
pdp2121: Combined them and tested locally.

src/shared/database/setup.ts
```diff
@@ -179,15 +179,21 @@ async function setupNetworksTable(): Promise<void> {
       table.string('unls')
       table.primary(['entry'])
     })
+  }
+  const networksCount = await query('networks').countDistinct('id')
+  const networksIds = await query('networks').pluck('id')
+  if (networksCount[0].count < networks.length) {
```
mvadari, line 185: What's the objective of this if statement?

src/api/routes/v1/validator.ts
pdp2121: I reverted to old method. Not sure how to clean up on production yet, probably will reach out to you tmr for instructions.

src/shared/database/setup.ts
pdp2121: Good catch. Yea we dont need it after combined.

ckniffen (approved):

mvadari (approved):

src/api/routes/v1/validator.ts
pdp2121: @ckniffen

PR #52, opened 2022-10-13
Use UNL to retrieve manifest

pdp2121:

The xrpl web socket (http://xrpl.ws/) is only for mainnet and nodes from other networks are hit because the nodes are interconnecting. Pull manifests from individual unl using http request instead (see https://xrpl.org/validator-list.html#request-format) to yield more correct result.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Manifests for validators on the list of networks would be updated. API calls to retrieve validators of chains such as amm that are not interconnect with mainnet would yield a non-empty list of validators.

### DIFF
--- a/src/connection-manager/manifests.ts
+++ b/src/connection-manager/manifests.ts
@@ -3,13 +3,7 @@ import {
   verifyValidatorDomain,
 } from 'xrpl-validator-domains'
 
-import {
-  saveManifest,
-  getValidatorSigningKeys,
-  query,
-  db,
-  getNetworks,
-} from '../shared/database'
+import { saveManifest, query, db, getNetworks } from '../shared/database'
 import {
   StreamManifest,
   Manifest,
@@ -17,7 +11,7 @@ import {
   UNLValidator,
   DatabaseManifest,
 } from '../shared/types'
-import { fetchValidatorList, fetchRpcManifest, getLists } from '../shared/utils'
+import { fetchValidatorList, getLists } from '../shared/utils'
 import logger from '../shared/utils/logger'
 
 import hard_dunl from './fixtures/unl-hard.json'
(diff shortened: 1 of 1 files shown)


test/manifests/manifests.test.ts
```diff
@@ -92,56 +90,6 @@ describe('manifest ingest', () => {
     })
   })
 
-  test('updateManifestsFromRippled', async () => {
```
mvadari, line 95: This test case should be replaced with what it's doing now

src/connection-manager/manifests.ts
```diff
@@ -85,54 +79,24 @@ export async function handleManifest(
 export async function updateUNLManifests(): Promise<void> {
   try {
     log.info('Fetching UNL...')
-    const unl: UNLBlob = await fetchValidatorList(await getFirstUNL('main'))
```
mvadari, line 88: The docstring from this method should be updated to be more clear

src/connection-manager/manifests.ts
```diff
@@ -85,54 +79,24 @@ export async function handleManifest(
...
   }
 }
 
-/**
- * This function loops through all signing keys in the validators table and queries rippled
- * to find the most recent manifest available.
- *
- * @returns A promise that resolves to void once all of the latest manifests have been saved.
- */
-export async function updateManifestsFromRippled(): Promise<void> {
```
mvadari, line 109: This might still be useful to do for mainnet

pdp2121: Validators on mainnet is updated with xl.ripple.com unl. If you want to keep those unused ones then sure I can keep this function.

test/manifests/manifests.test.ts
pdp2121: Yes, the new manifests retrieval is tested with the `update Manifests` test.

src/connection-manager/manifests.ts
```diff
@@ -84,55 +80,24 @@ export async function handleManifest(
...
-    const promises: Array<Promise<void>> = []
+    log.info('Fetching Manifests from UNLs...')
+    const networks = await getNetworks()
 
-    unl.validators.forEach((validator: UNLValidator) => {
-      const manifestHex = Buffer.from(validator.manifest, 'base64')
-        .toString('hex')
-        .toUpperCase()
-      promises.push(handleManifest(manifestHex))
+    await async.map(networks, async (network: Network) => {
```
mvadari, line 86: What does `async` do here?

test/manifests/manifests.test.ts
```diff
@@ -73,76 +71,16 @@ describe('manifest ingest', () => {
...
-        '304402200A3F6ECD0218C24D94CF12892F6BF41BFD1B7E58793D05DB41A1A27672908C1102206221014E7FA98D5BE43203C00D994689F07C23B345518A3297BD9AC70EA00C84',
-      revoked: false,
-      domain: null,
-      domain_verified: false,
-      seq: '8',
-    })
+    const saved_manifest = await query('manifests')
+      .select('*')
+      .orderBy('master_key')
+    expect(saved_manifest.length).toBeGreaterThan(0)
```
mvadari, line 80: This should check what the result is, instead of just asserting that it exists

src/connection-manager/manifests.ts
mvadari: I'm thinking that it would be useful for the case where a validator's manifest changes but the UNL hasn't been updated yet.

mvadari: If you look here: https://livenet.xrpl.org/network/validators the domain is available for a bunch of validators. If we just pull from the UNL, we won't have the domains for the other validators. This makes it hard to e.g. identify validator operators that need to upgrade

pdp2121: I see. Will restore the functionality then.

src/connection-manager/manifests.ts
pdp2121: `async` would be to ensure all inner promises got executed before the function returns, since this function incurs multiple loop layers of async execution. It could be done as well with multiple Promise.all layers but it would raise eslint issues and also returns a `Promise<void[][]>`. @ckniffen and I think it would be cleaner this way using async package.

test/manifests/manifests.test.ts
pdp2121: The manifests data would change overtime so I'm not sure how to make an appropriate test. Any suggestions?

mvadari: You can mock it, like other tests do.

src/connection-manager/manifests.ts
```diff
@@ -1,23 +1,25 @@
+import async from 'async'
 import {
   normalizeManifest,
   verifyValidatorDomain,
 } from 'xrpl-validator-domains'
 
 import {
   saveManifest,
-  getValidatorSigningKeys,
```
mvadari, line 8: Why is the linting changing here?

test/manifests/manifests.test.ts
```diff
@@ -73,22 +73,23 @@ describe('manifest ingest', () => {
     })
   })
 
-  test('updateUnlManifests', async () => {
```
mvadari, line 76: Why were these tests changed so much?

src/connection-manager/manifests.ts
```diff
@@ -84,17 +86,19 @@ export async function handleManifest(
...
+      const unl: UNLBlob = await fetchValidatorList(network.unls[0])
+      await async.map(unl.validators, async (validator: UNLValidator) => {
+        const manifestHex = Buffer.from(validator.manifest, 'base64')
+          .toString('hex')
+          .toUpperCase()
+        return handleManifest(manifestHex)
+      })
     })
-    await Promise.all(promises)
+    return
```
mvadari, line 101: This `return` is unnecessary

test/manifests/manifests.test.ts
ckniffen: I would think the same keys would work in the testcase.

pdp2121: The previous updateUNLManifests just get manifests from `xl.ripple.com`, while the new function get manifests from all of them, which would yield different results.

pdp2121: `const unl: UNLBlob = await fetchValidatorList(await getFirstUNL('main'))`. This is what in the function before the change

pdp2121: I fixed the test so that it could use the mock data and same keys

mvadari:
Closed in favor of #57

PR #53, opened 2022-10-17
fix: crawls networks to use most updated network.

pdp2121:

Currently, the networks column of crawls table is a list of networks that updates when a node changes its network by adding a comma and a network name to the string. This PR will update the node with just the most recent network.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### Before

When a node in `test` changes to `dev`, the column `networks` in the `crawls` table will show `test,dev`.

### After
When a node in `test` changes to `dev`, the column `networks` in the `crawls` table will just show `dev`.


ckniffen (approved):

mvadari:
I'd like this to run in staging for a couple weeks before I'm comfortable approving it. 

I have no doubt that this change will do what we're expecting it to do, I'm just not 100% sure that's what we want to do.

intelliot (approved):
if it doesn't work, you can revert it

PR #54, opened 2022-10-24
Add networks column to Validators table

pdp2121:

Currently, the validator API endpoint in the VHS uses the UNL URL to determine which validators to return, but that means that a network needs a UNL to be tracked. Sidechains/other custom networks don't necessarily use UNLs - they may have a hardcoded list of servers or something.

To accommodate for this, we could add a "network" column to the "validators" table to indicate which network the validator came from. This could be derived from the node whose WS port the VHS gets the service from (basically, the VHS subscribes to a bunch of nodes and listens to their validations streams).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

A new column `networks` would be added to `validators` table.
A new API endpoint `network/validators/{networks}` will be added.


src/shared/types/index.ts
```diff
@@ -17,6 +17,18 @@ interface Ledger {
   first_seen: number
 }
 
+interface StreamLedger {
```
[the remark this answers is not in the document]
pdp2121: We deleted the `ledgers` table but still stream `ledgers` to get the hashes so that's what this interface is for.

src/connection-manager/connections.ts
```diff
@@ -25,19 +27,53 @@ function subscribe(ws: WebSocket): void {
...
+ */
+async function handleWsMessageTypes(
+  data: ValidationRaw | StreamManifest | StreamLedger,
+  ledger_hashes: string[],
+  networks: string | undefined,
+): Promise<void> {
+  if (data.type === 'validationReceived') {
+    if (ledger_hashes.includes((data as ValidationRaw).ledger_hash)) {
+      ;(data as ValidationRaw).networks = networks
+    }
```
mvadari, line 51:
```suggestion
    const validationData = data as ValidationRaw
    if (ledger_hashes.includes((validationData).ledger_hash)) {
      validationData.networks = networks
    }
```

intelliot (approved):
LGTM (I didn't try it)

src/api/routes/v1/info.ts
```diff
@@ -31,6 +31,11 @@ const info = {
       route: '/v2/network/validators/{unl}',
       example: 'https://data.xrpl.org/v1/network/validators/{unl}',
     },
+    {
+      action: 'Get Validators By Networks',
+      route: '/v2/network/validators/{networks}',
+      example: 'https://data.xrpl.org/v1/network/validators/{networks}',
```
mvadari, line 37: the version is different between these two

src/api/routes/v1/validator.ts
```diff
@@ -259,19 +264,27 @@ export async function handleValidator(
...
     return undefined
   }
+  const networksDb = await getNetworks()
+  const networks = networksDb.map((network) => network.id)
+  let requestedField
+  if (networks.includes(param)) {
+    requestedField = 'networks'
+  } else {
+    requestedField = 'unl'
+  }
```
mvadari, line 283: This should also check that the param is a valid URL, in this case

ckniffen (approved):

mvadari (approved):

PR #59, opened 2022-11-23
subscribe initial nodes to the streams

pdp2121:

Add initial entry nodes of each network to the streams, in case other nodes in the network does not have an ip and ws_url.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


mvadari (approved):

src/connection-manager/connections.ts
```diff
@@ -167,6 +172,15 @@ async function createConnections(): Promise<void> {
     .whereNotNull('ip')
     .andWhere('start', '>', tenMinutesAgo)
 
+  const networksDb = await getNetworks()
+  networksDb.forEach((network) => {
+    nodes.push({
+      ip: network.entry,
+      ws: `ws://${network.entry}:51233`,
```
mvadari, line 179: You might want to add a port check in this, in case a port is mentioned

pdp2121: The port in db is 51235, which is not what we want here. The findConnection function actually tried all ports in this list `[443, 80, 6005, 6006, 51233, 51234]` if we leave ws blank (line 149-155 of this file). Should I leave ws blank then?

src/connection-manager/connections.ts
```diff
@@ -167,6 +172,15 @@ async function createConnections(): Promise<void> {
     .whereNotNull('ip')
     .andWhere('start', '>', tenMinutesAgo)
 
+  const networksDb = await getNetworks()
+  networksDb.forEach((network) => {
+    nodes.push({
+      ip: network.entry,
+      ws: '',
```
mvadari, line 179:
```suggestion
      ws_url: '',
```

mvadari (approved):

ckniffen (approved):

PR #60, opened 2022-12-06
feat: add UNL to `get_network` API

mvadari:

Title says it all. This PR also documents the `get_network` API.

### Context of Change

This is necessary for https://github.com/ripple/explorer/pull/470

### Type of Change

## Test Plan

Works locally.

### DIFF
--- a/src/api/routes/v1/get-network.ts
+++ b/src/api/routes/v1/get-network.ts
@@ -124,6 +124,7 @@ async function any(
  */
 async function fetchCrawls(host: string): Promise<Crawl> {
   const promises: Array<Promise<Crawl | undefined>> = []
+  log.info(`crawling ${host}...`)
   for (const port of CRAWL_PORTS) {
     promises.push(crawlNode(host, port))
   }
@@ -193,6 +194,7 @@ async function addNode(url: string, unl: string | null): Promise<string> {
     port: 51235,
     unls: unl ? [unl] : [],
   }
+  log.info(`adding new network ${JSON.stringify(network)}`)
   await query('networks').insert({
     ...network,
     unls: network.unls.join(','),
--- a/src/api/routes/v1/get-network.ts
+++ b/src/api/routes/v1/get-network.ts
@@ -11,6 +11,11 @@ const log = logger({ name: 'api-get-network' })
 
 const CRAWL_PORTS = [51235, 2459, 30001]
 
+interface NetworkInfo {
+  network: string
+  unl?: string[]
+}
+
 let maxNetwork: number
 
 async function updateMaxNetwork(): Promise<void> {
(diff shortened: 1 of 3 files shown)


pdp2121 (approved):

src/api/routes/v1/info.ts
```diff
@@ -6,6 +6,11 @@ const info = {
   documentation: 'Put Docs HERE',
   release_notes: 'Release Notes HERE',
   endpoints: [
+    {
+      action: 'Get Network From Node',
+      route: '/v1/network/get_network',
+      example: 'https://data.xrpl.org/v1/network/get_network',
```
ckniffen, line 12: The example should include that it takes a parameter.

src/api/routes/v1/get-network.ts
```diff
@@ -11,6 +11,11 @@ const log = logger({ name: 'api-get-network' })
 
 const CRAWL_PORTS = [51235, 2459, 30001]
 
+interface NetworkInfo {
+  network: string
+  unl?: string[]
```
pdp2121, line 16: Should it just be a string since it only takes the first unl?

src/api/routes/v1/get-network.ts
```diff
@@ -147,13 +161,20 @@ async function getNetworkFromUNL(unl: string): Promise<string | undefined> {
...
     .select('public_key', 'networks')
     .where('public_key', '=', publicKey)
-  return result.length > 0 ? result[0].networks : undefined
+  const networkIds = result.length > 0 ? result[0].networks : undefined
+  if (networkIds == null) {
+    return networkIds
+  }
+  const networkResult = await query('networks')
+    .select('id', 'unls')
+    .whereIn('id', networkIds.split(','))
```
pdp2121, line 176: This is a bit different than result from `getNetworkInfoFromUNL` above since field name would be unls instead of unl and would be a string of type '...,...,...`.

src/api/routes/v1/get-network.ts
```diff
@@ -132,11 +138,19 @@ async function fetchCrawls(host: string): Promise<Crawl> {
...
+): Promise<NetworkInfo | undefined> {
   const result = await query('networks')
-    .select('id')
+    .select('id', 'unls')
     .where('unls', 'like', `%${unl}%`)
-  return result.length > 0 ? result[0].id : undefined
+  if (result.length > 0) {
+    return {
+      network: result[0].id,
+      unl: result[0].unls.split(',')[0],
```
pdp2121, line 150: Should we just show all unls similar to `getNetworkInfoFromPublicKey` below?

intelliot:
@mvadari is this PR still needed?

src/api/routes/v1/get-network.ts
```diff
@@ -147,13 +161,20 @@ async function getNetworkFromUNL(unl: string): Promise<string | undefined> {
...
     .where('public_key', '=', publicKey)
-  return result.length > 0 ? result[0].networks : undefined
+  const networkIds = result.length > 0 ? result[0].networks : undefined
+  if (networkIds == null) {
+    return networkIds
+  }
+  const networkResult = await query('networks')
+    .select('id', 'unls')
+    .whereIn('id', networkIds.split(','))
+  return networkResult
```
ckeshava, line 177: perhaps, we should add a length check on the return of this query (or) return the `[0]`'th element. It appears that the comma-separated `networkIds` list could contain multiple elements.

src/api/routes/v1/get-network.ts
```diff
@@ -132,11 +138,19 @@ async function fetchCrawls(host: string): Promise<Crawl> {
  * @param unl - The UNL of the node.
  * @returns Whether the UNL of the node has been seen before.
  */
-async function getNetworkFromUNL(unl: string): Promise<string | undefined> {
+async function getNetworkInfoFromUNL(
```
ckeshava, line 141: This API endpoint has not been documented in the `ARCHITECTURE.md` file. How am I supposed to use it?

This [URL](https://data.xrpl.org/v1/network/get_network/s2.ripple.com) does not work

image

PR #75, opened 2023-02-09
fix: UNLs does not get updated when a network has empty UNLs

pdp2121:

`updateUNLs` is failing for all validators on staging and production since there's a new network added with empty UNLs. Prevent this from happening by fixing `Promise.all` in `fetchValidatorList` to continue even when a promise raises an error.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Ensure `updateUNLs` run successfully in staging and validator with pubkey `nHUon2tpyJEHHYGmxqeGu37cvPYHzrMtUNQFVdCgGNvEkjmCpTqK` is removed from vl.ripple.com.

### DIFF
--- a/src/shared/utils/index.ts
+++ b/src/shared/utils/index.ts
@@ -27,7 +27,10 @@ export async function fetchValidatorList(
     const blobParsed: UNLBlob = JSON.parse(buf.toString('ascii'))
     return blobParsed
   } catch (err: unknown) {
-    log.error(`Error fetching validator List for network ${networkId}`, err)
+    log.error(
+      `Error fetching validator List for network ${networkId} and url ${url}`,
+      err,
+    )
     return Promise.reject()
   }
 }
@@ -123,13 +126,15 @@ export async function getLists(): Promise<Record<string, Set<string>>> {
   const promises: Array<Promise<void>> = []
   const networks = await getNetworks()
   networks.forEach(async (network) => {
-    promises.push(
-      fetchValidatorList(network.unls[0], network.id).then((blob) => {
-        Object.assign(lists, {
-          [network.id]: blobToValidators(blob),
-        })
-      }),
-    )
+    if (network.unls[0]) {
+      promises.push(
+        fetchValidatorList(network.unls[0], network.id).then((blob) => {
+          Object.assign(lists, {
+            [network.id]: blobToValidators(blob),
+          })
+        }),
+      )
+    }
   })
   await Promise.all(
     // The error has already been logged in fetchValidatorList
--- a/src/shared/utils/index.ts
+++ b/src/shared/utils/index.ts
@@ -127,7 +127,9 @@ export async function getLists(): Promise<Record<string, Set<string>>> {
       }),
     )
   })
-  await Promise.all(promises)
+  await Promise.all(
+    promises.map(async (promise) => promise.catch(async (err) => err)),
+  )
   return lists
 }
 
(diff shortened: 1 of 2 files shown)


src/shared/utils/index.ts
```diff
@@ -127,7 +127,9 @@ export async function getLists(): Promise<Record<string, Set<string>>> {
       }),
     )
   })
-  await Promise.all(promises)
+  await Promise.all(
+    promises.map(async (promise) => promise.catch(async (err) => err)),
```
mvadari, line 131: The error should be logged.

pdp2121: The error will still be logged as fetchValidatorList ran for the empty network.

pdp2121: Screenshot 2023-02-09 at 3 39 20 PM

pdp2121: I will add to the error message to display network id

ckniffen: If you catch and it returns this way the promise will resolve with the error.  Not a fan of that.  You should log it instead.  Additionally the async commands are not needed.

pdp2121: Discussed with Caleb and added a comment to indicate that the error has been logged in `fetchValidatorList`

mvadari (approved):

src/shared/utils/index.ts
```diff
@@ -13,17 +13,21 @@ const HTTPS_PORT = 51234
...
+): Promise<UNLBlob> {
   try {
     const resp = await axios.get(`http://${url}`)
     const unl: UNL = resp.data
     const buf = Buffer.from(unl.blob, 'base64')
     const blobParsed: UNLBlob = JSON.parse(buf.toString('ascii'))
     return blobParsed
   } catch (err: unknown) {
-    log.error('Error fetching validator List', err)
+    log.error(`Error fetching validator List for network ${networkId}`, err)
```
ckniffen, line 30: Looking at this I think you could just pass in the url instead of the network id.  That would reduce passing in a variable just for error messaging.

src/shared/utils/index.ts
```diff
@@ -119,15 +126,20 @@ export async function getLists(): Promise<Record<string, Set<string>>> {
...
   const networks = await getNetworks()
   networks.forEach(async (network) => {
-    promises.push(
-      fetchValidatorList(network.unls[0]).then((blob) => {
-        Object.assign(lists, {
-          [network.id]: blobToValidators(blob),
-        })
-      }),
-    )
+    if (network.unls[0]) {
```
ckniffen, line 129: instead of hiding it hear return early.  That will unindent more relavant code.

`if(!network.unls[0]){ return }`

src/shared/utils/index.ts
```diff
@@ -13,17 +13,24 @@ const HTTPS_PORT = 51234
...
+): Promise<UNLBlob> {
   try {
     const resp = await axios.get(`http://${url}`)
     const unl: UNL = resp.data
     const buf = Buffer.from(unl.blob, 'base64')
     const blobParsed: UNLBlob = JSON.parse(buf.toString('ascii'))
     return blobParsed
   } catch (err: unknown) {
-    log.error('Error fetching validator List', err)
+    log.error(
```
ckniffen, line 30: network id not needed as url will always be set.

src/shared/utils/index.ts
ckniffen: Don't return the error.

ckniffen (approved):

PR #76, opened 2023-02-15
fix: use most updated network for validators

pdp2121:

`Validators` table is updating networks with comma-separated. This change will keep only the most updated one.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -9,7 +9,6 @@ import {
   purgeHourlyAgreementScores,
   signingToMaster,
   decodeServerVersion,
-  query,
 } from '../shared/database'
 import {
   AgreementScore,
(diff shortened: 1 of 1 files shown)


mvadari:
lgtm, assuming it fixes the issue in staging - holding off on ✅ until then

ckniffen:
Is #53 also needed?  I think it makes sense to combine with it.

pdp2121:
> Is #53 also needed? I think it makes sense to combine with it.

Yes I think it makes sense to just keep one network for both nodes and validators. @mvadari what do you think?

mvadari (approved):

mvadari:
> > Is #53 also needed? I think it makes sense to combine with it.
> 
> Yes I think it makes sense to just keep one network for both nodes and validators. @mvadari what do you think?

I agree - we should just use the same policy for both. Though I do think it'll be easier to test in prod if we only use this policy for one of the two for a couple months, just to see how it looks.

pdp2121:
> > > Is #53 also needed? I think it makes sense to combine with it.
> > 
> > 
> > Yes I think it makes sense to just keep one network for both nodes and validators. @mvadari what do you think?
> 
> I agree - we should just use the same policy for both. Though I do think it'll be easier to test in prod if we only use this policy for one of the two for a couple months, just to see how it looks.

Okay I will deploy this first and hold off on #53 for like a month to see how it goes.

ckniffen (approved):

PR #77, opened 2023-02-16
fix: remove nft-dev network

pdp2121:

Title said it all.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/shared/database/setup.ts
+++ b/src/shared/database/setup.ts
@@ -185,4 +185,10 @@ async function setupNetworksTable(): Promise<void> {
         .catch((err: Error) => log.error(err.message))
     }
   })
+  if (networksIds.includes('nft-dev')) {
+    query('networks')
+      .del()
+      .where('id', '=', 'nft-dev')
+      .catch((err: Error) => log.error(err.message))
+  }
 }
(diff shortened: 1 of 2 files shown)


mvadari (approved):

ckniffen (approved):

src/shared/database/setup.ts
```diff
@@ -185,4 +185,10 @@ async function setupNetworksTable(): Promise<void> {
         .catch((err: Error) => log.error(err.message))
     }
   })
+  if (networksIds.includes('nft-dev')) {
```
ckniffen, line 188: Will this cause foreign key errors?

pdp2121: It doesnt use a foreign key. It just check whether nft-dev is in the array of ids of current networks on the table and if so delete that row using knex.

ckniffen:
You should combine with #78

src/shared/database/setup.ts
mvadari: I don't think the networks table is used as a foreign key anywhere

mvadari (approved):

PR #94, opened 2023-06-20
feat: add hooks testnet support

mvadari:

Title says it all.

### Context of Change

Necessary to add support for the hooks testnet to the Explorer.

### Type of Change

## Test Plan

Works locally.


pdp2121 (approved):

src/shared/database/networks.ts
```diff
@@ -46,6 +46,12 @@ const networks: Network[] = [
     port: 51235,
     unls: ['vlamm.devnet.rippletest.net'],
   },
+  {
+    id: 'hooks-test',
+    entry: '88.99.3.241',
```
ckniffen, line 51: Why are you referencing an IP address instead of the uri?

mvadari: The URL doesn't have an open crawl.

I went through the public peers listed here: https://xumm.notion.site/Hooks-V3-staging-net-info-518fa261c5cd49d2bcb89a5b9e7bef05
Two of the IPs had public crawling so I had to use one of those.

ckniffen: Will it still recognize the network when sending the url version to the vhs as the entry point?

mvadari: Yes, via the ID.

mvadari: This is why I want network ID support for the VHS - it simplifies that logic.

ckniffen (approved):

PR #101, opened 2023-07-12
feat: add validators voting data to db

pdp2121:

On every most recent flag ledger, store for each validator into `ballot` table:

- ledger index - this will allow us to keep the most recent vote for a validator incase we missed the most recent flag ledger.

- fee voting

- amendment voting

This is to get amendment voting status and later fee votes.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### Future Tasks
- Add voting data to `validators` endpoint.
- Add a new `amendments` endpoint to returns amendments enabled on a network and voting data for pending ones.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -113,7 +113,6 @@ This table keeps track of the most current voting data for the validators.
 
 | Key                  | Definition                                                        |
 |----------------------|-------------------------------------------------------------------|
-| `master_key`         |The master key of the validator.                                   |
 | `signing_key`        |The signing key of the validator.                                  |
 | `ledger_index`       |The most recent ledger index where voting data was retrieved.      |
 | `amendments`         |The amendments this validator wants to be added to the protocol.   |
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -107,6 +107,21 @@ This table keeps track of the manifests of the validators.
 | `seq`                |The sequence number of this manifest.                       |
 
 
+### `ballot`
+
+This table keeps track of the most current voting data for the validators.
+
+| Key                  | Definition                                                        |
+|----------------------|-------------------------------------------------------------------|
+| `master_key`         |The master key of the validator.                                   |
+| `signing_key`        |The signing key of the validator.                                  |
+| `ledger_index`       |The most recent ledger index where voting data was retrieved.      |
+| `amendments`         |The amendments this validator wants to be added to the protocol.   |
+| `base_fee`           |The unscaled transaction cost this validator wants to set.         |
+| `reserve_base`       |The minimum reserve requirement this validator wants to set.       |
+| `reserve_inc`        |Whether the validation is a partial validation*.                   |
+
+
 ### `validators`
 
 This table keeps track of the validators on the networks.
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -119,7 +119,7 @@ This table keeps track of the most current voting data for the validators.
 | `amendments`         |The amendments this validator wants to be added to the protocol.   |
 | `base_fee`           |The unscaled transaction cost this validator wants to set.         |
 | `reserve_base`       |The minimum reserve requirement this validator wants to set.       |
-| `reserve_inc`        |Whether the validation is a partial validation*.                   |
+| `reserve_inc`        |The increment in the reserve requirement this validator wants to set.|
 
 
 ### `validators`
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -202,7 +202,7 @@ class Agreement {
           master_key: validation.master_key,
           signing_key,
           ledger_index: Number(validation.ledger_index),
-          amendments: validation.amendments?.toString(),
+          amendments: validation.amendments?.join(','),
           base_fee: validation.base_fee,
           reserve_base: validation.reserve_base,
           reserve_inc: validation.reserve_inc,
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -9,12 +9,14 @@ import {
   purgeHourlyAgreementScores,
   signingToMaster,
   decodeServerVersion,
+  saveBallot,
 } from '../shared/database'
 import {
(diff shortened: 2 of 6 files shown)


pdp2121:
I will continue to monitor the code locally (looked good till now). Pushed this to get some feedback.

ARCHITECTURE.md
```diff
@@ -107,6 +107,21 @@ This table keeps track of the manifests of the validators.
 | `seq`                |The sequence number of this manifest.                       |
 
 
+### `ballot`
```
mvadari, line 110: Thanks for keeping this doc updated 🙂

src/connection-manager/agreement.ts
```diff
@@ -195,6 +194,22 @@ class Agreement {
...
 
+      let server_version
+
+      if (isPreceedingFlagLedger(validation.ledger_index)) {
+        server_version = decodeServerVersion(validation.server_version)
+        const ballot: Ballot = {
+          master_key: validation.master_key,
+          signing_key,
+          ledger_index: Number(validation.ledger_index),
+          amendments: validation.amendments?.toString(),
```
mvadari, line 205: This should be a comma-separated list or something, for easier parsing

mvadari (commented):
There needs to be an API endpoint to access this data

pdp2121:
> There needs to be an API endpoint to access this data

Yes, that would be the upcoming PR. We separated this into smaller tasks

src/connection-manager/agreement.ts
pdp2121: Yes it is comma-separated

mvadari: Really? Usually `list.toString()` generates something more like `[1, 2, 3]` instead of `1,2,3`

pdp2121: `27CD95EE8E1E5A537FF2F89B6CEB7C622E78E9374EBD7DCBEDFAE21CD6F16E0A,56B241D7A43D40354D02A9DC4C8DF5C7A1F930D92A9035C4E12291B3CA3E1C2B` this is one example from db, but sure I can use `join` to make sure it's always correct

mvadari: Yeah, I'd prefer a join.
Will there be any issues with string lengths here?

pdp2121: Yea that's the part I'm uncertain about, I set the length to be 10,000 for now, but not sure if there'll ever be a case when it get overflow

mvadari: That seems pretty reasonable for now, that takes 156 amendments actively being voted on which is unlikely

intelliot:
> > There needs to be an API endpoint to access this data
> 
> Yes, that would be the upcoming PR. We separated this into smaller tasks

That's good, but you should note this in the "Future Tasks" section of the PR description template.

pdp2121:
> > > There needs to be an API endpoint to access this data
> > 
> > 
> > Yes, that would be the upcoming PR. We separated this into smaller tasks
> 
> That's good, but you should note this in the "Future Tasks" section of the PR description template.

I have updated the description to include future tasks. Thanks Elliot!

ARCHITECTURE.md
```diff
@@ -107,6 +107,21 @@ This table keeps track of the manifests of the validators.
...
+
+This table keeps track of the most current voting data for the validators.
+
+| Key                  | Definition                                                        |
+|----------------------|-------------------------------------------------------------------|
+| `master_key`         |The master key of the validator.                                   |
+| `signing_key`        |The signing key of the validator.                                  |
+| `ledger_index`       |The most recent ledger index where voting data was retrieved.      |
+| `amendments`         |The amendments this validator wants to be added to the protocol.   |
```
ckniffen, line 119: This might make sense to join another table to make it easier to query an amendment and get the list of validators voting on it.

src/connection-manager/agreement.ts
```diff
@@ -195,6 +194,22 @@ class Agreement {
...
+      let server_version
+
+      if (isPreceedingFlagLedger(validation.ledger_index)) {
+        server_version = decodeServerVersion(validation.server_version)
+        const ballot: Ballot = {
+          master_key: validation.master_key,
+          signing_key,
+          ledger_index: Number(validation.ledger_index),
+          amendments: validation.amendments?.join(','),
+          base_fee: validation.base_fee,
```
ckniffen, line 206: Should this store the current amount for this network when the validation does not contain these properties?

pdp2121: Could you explain what you meant by current amount?

ckniffen: If they dont vote they are saying they want it to be what the current rate is.  This could be populated here or derived when you ask for the amendment ballots.

ARCHITECTURE.md
```diff
@@ -107,6 +107,21 @@ This table keeps track of the manifests of the validators.
 | `seq`                |The sequence number of this manifest.                       |
 
 
+### `ballot`
+
+This table keeps track of the most current voting data for the validators.
+
+| Key                  | Definition                                                        |
+|----------------------|-------------------------------------------------------------------|
+| `master_key`         |The master key of the validator.                                   |
```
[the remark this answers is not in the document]
pdp2121: Yes we can just store `signing_key` to join with `validators` table

ARCHITECTURE.md
pdp2121: Yes we will need to join with the `validators` table for both `validators` and `amendments` APIs (for the second API we will need `networks` data from `validators` table)

src/connection-manager/agreement.ts
pdp2121: I see. Yes we can fill in those current rates using data from `ledger` stream

pdp2121:
@mvadari @ckniffen I have added fee data for validators that do not vote for a fee changed in the latest commit. Please let me know if it looks good, then we can test on staging

ARCHITECTURE.md
ckniffen: Will it be easier to do that with amendments just being an array of comma separated fields?

PR #102, opened 2023-07-18
chore: run tests on Node 18.x and 20.x

mvadari:

Title says it all.

### Context of Change

With Node 14.x being deprecated, I wanted to make sure the VHS ran on newer versions of Node

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -2,18 +2,17 @@ name: Node.js CI
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
   workflow_dispatch:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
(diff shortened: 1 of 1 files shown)


pdp2121 (approved):

.github/workflows/node.js.yml
```diff
@@ -2,18 +2,17 @@ name: Node.js CI
...
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [14.x, 16.x, 18.x, 20.x]
```
ckniffen, line 15: Is there a benefit on running this on versions we dont use ourselves?

ckniffen (approved):

.github/workflows/node.js.yml
mvadari: Other people can run it on other versions, and we know it'll work when we want to switch.

PR #104, opened 2023-07-27
feat: Add amendments to validators endpoint

pdp2121:

Add voting data to validators related endpoints.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

Ready to review when #101 merged.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -113,7 +113,6 @@ This table keeps track of the most current voting data for the validators.
 
 | Key                  | Definition                                                        |
 |----------------------|-------------------------------------------------------------------|
-| `master_key`         |The master key of the validator.                                   |
 | `signing_key`        |The signing key of the validator.                                  |
 | `ledger_index`       |The most recent ledger index where voting data was retrieved.      |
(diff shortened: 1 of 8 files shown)


ckniffen:
We also need an overall voting information and to know which amendments are enabled.

Additionally a look up for information on amendments might be very nice.  Right now the explorer has weird logic where it parses the documentation site and a rippled cpp file.

ckniffen:
Here is where that happens. https://github.com/ripple/explorer/blob/staging/src/containers/shared/amendmentUtils.ts#L32-L49 and https://github.com/ripple/explorer/blob/staging/src/containers/shared/amendmentUtils.ts#L72-L90

pdp2121:
> Here is where that happens. https://github.com/ripple/explorer/blob/staging/src/containers/shared/amendmentUtils.ts#L32-L49 and https://github.com/ripple/explorer/blob/staging/src/containers/shared/amendmentUtils.ts#L72-L90

Thanks Caleb, I was also thinking about parsing the doc site for amendment info, but not sure about rippled file. Will examine these codes

PR #117, opened 2023-08-11
feat: add amendments enabled on each network

pdp2121:

Create a table with the following information to track amendments enabled on each network:

- AmendmentID

- NetworkID

- ledger_index

- tx_hash

- date

Also add a function to manually update the table using a JSON file (runs whenever re-deployed).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Works on staging.


src/connection-manager/connections.ts
```diff
@@ -37,6 +45,21 @@ function subscribe(ws: WebSocket): void {
   )
 }
 
+/**
+ * Sends a ledger_entry WebSocket request to retrieve amendments enabled on a network.
+ *
+ * @param ws - A WebSocket object.
+ */
+function ledger_entry(ws: WebSocket): void {
```
mvadari, line 53: This shouldn't be called just `ledger_entry`, it should be something like `getAmendmentLedgerEntry` - it's not just a wrapper for _any_ `ledger_entry` request.

src/shared/types/index.ts
```diff
@@ -29,6 +29,37 @@ interface StreamLedger {
...
+}
+
+interface LedgerEntryMajority {
+  Majority: {
+    Amendment: string
+    CloseTime: number
+  }
+}
+
+interface LedgerEntryAmendments {
```
mvadari, line 44: nit:
```suggestion
interface LedgerEntryAmendmentsResponse {
```

src/connection-manager/connections.ts
```diff
@@ -89,6 +123,9 @@ async function setHandlers(
       void saveNodeWsUrl(ws.url, true)
       connections.set(ip, ws)
       subscribe(ws)
+      if (entry) {
```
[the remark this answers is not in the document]
pdp2121: We just need to call `ledger_entry` once per network to get currently enabled amendments list so I just call it using entry nodes

src/connection-manager/connections.ts
```diff
@@ -6,13 +8,18 @@ import {
   clearConnectionsDb,
   getNetworks,
   saveAmendmentsEnabled,
+  saveAmendmentEnabled,
```
mvadari: Why are these two separate functions?

mvadari (commented):
For ease of reviewing, please set up the PRs so they merge one into another. You have write access on the repo so you can create + push branches directly off of the main repo instead of off of a fork.

PR #118, opened 2023-08-16
feat: add `amendments/vote` endpoint

pdp2121:

Add an api endpoint to retrieve lists of enabled and voting amendments for a network.
This PR will combine all existing PR to test on staging.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Test on staging for a few days.


ARCHITECTURE.md
```diff
@@ -23,6 +23,8 @@ There are 3 folders in `src`, corresponding to the 3 processes that the VHS runs
   * `/network/validator/:publicKey`: Returns information about a specific validator.
   * `/network/validator/:publicKey/manifests`: Returns the manifests of a specific validator.
   * `/network/validator/:publicKey/reports`: Returns more detailed information about the reliability of a specific validator.
+  * `/network/amendments/info`: Returns general information about known amendments.
```
mvadari, line 26: Can this API endpoint also have a version for information about a specific amendment (with support for both the hash and the name, ideally)?

ARCHITECTURE.md
```diff
@@ -106,6 +108,38 @@ This table keeps track of the manifests of the validators.
 | `revoked`            |Whether the manifest has been revoked.                      |
 | `seq`                |The sequence number of this manifest.                       |
 
+### `ballot`
```
mvadari, line 111: IMO this should be `amendment_ballot` or something - `ballot` is a bit vague

src/api/routes/v1/amendments.ts
```diff
@@ -0,0 +1,315 @@
...
+ * @param id - The network id.
+ * @returns List of enabled amendments.
+ */
+async function getVotingAmendments(id: string): Promise<AmendmentInVoting[]> {
+  const inNetworks: BallotAmendmentDb[] = await query('ballot')
+    .leftJoin('validators', 'ballot.signing_key', 'validators.signing_key')
+    .select('ballot.signing_key', 'ballot.ledger_index', 'ballot.amendments')
+    .where('validators.networks', id)
+
+  const votingAmendments: {
```
mvadari, line 181: Why is this a map instead of just a list?

mvadari, line 181: Why is this a map instead of just a list?

src/api/routes/v1/index.ts
```diff
@@ -1,17 +1,22 @@
 import { Router as createRouter } from 'express'
 
+import { handleAmendmentsInfo, handleAmendmentsVote } from './amendments'
 import handleDailyScores from './daily-report'
 import getNetworkOrAdd from './get-network'
 import handleHealth from './health'
 import handleValidatorManifest from './manifests'
 import { handleNode, handleNodes, handleTopology } from './nodes'
 import { handleValidator, handleValidators } from './validator'
+// eslint-disable-next-line import/max-dependencies -- Disabled since this module requires multiple dependencies.
```
mvadari, line 10: This should be disabled for the file instead of just for the line

src/api/routes/v1/validator.ts
```diff
@@ -42,6 +42,10 @@ interface ValidatorResponse {
   } | null
   partial: boolean
   unl: boolean
```
mvadari, line 44: Good catch on updating this endpoint too

ARCHITECTURE.md
pdp2121: Maybe `validator_voting` since it also has fee votes not just amendments

src/api/routes/v1/amendments.ts
pdp2121: I used a map for quicker search when merging with `cacheInfo` and `ballot` data. There might be a more efficient way though but this is what I came up with

PR #130, opened 2023-09-20
feat: add validators voting data to db

pdp2121:

On every most recent flag ledger, store for each validator into `ballot` table:

- ledger index - this will allow us to keep the most recent vote for a validator incase we missed the most recent flag ledger.

- fee voting

- amendment voting

This is to get amendment voting status and later fee votes.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### Future Tasks
- Add voting data to `validators` endpoint.
- Add a new `amendments` endpoint to returns amendments enabled on a network and voting data for pending ones.


src/connection-manager/agreement.ts
```diff
@@ -195,6 +194,23 @@ class Agreement {
...
 
+      let server_version
+
+      if (isPreceedingFlagLedger(validation.ledger_index)) {
+        server_version = decodeServerVersion(validation.server_version)
+        const ballot: Ballot = {
+          signing_key,
+          ledger_index: Number(validation.ledger_index),
+          amendments: validation.amendments?.join(','),
+          base_fee: validation.base_fee ?? validation.ledger_fee?.fee_base,
```
mvadari, line 205: What is `validation.base_fee` as opposed to `validation.ledger_fee.fee_base`?

src/connection-manager/connections.ts
```diff
@@ -55,11 +62,31 @@ async function handleWsMessageTypes(
     if (ledger_hashes.includes(validationData.ledger_hash)) {
       validationData.networks = networks
     }
-    void agreement.handleValidation(data as ValidationRaw)
+    const validationNetworkDb: DatabaseValidator | undefined = await query(
```
mvadari, line 65: Can you comment the code in this file? It'll be easier to follow.

src/connection-manager/agreement.ts
```diff
@@ -195,6 +194,23 @@ class Agreement {
         last_ledger_time: new Date(),
       }
 
+      let server_version = null
```
mvadari, line 197: Can you comment the code in this file? It'll be easier to follow.

src/connection-manager/connections.ts
```diff
@@ -55,11 +62,31 @@ async function handleWsMessageTypes(
...
+      'validators',
+    )
+      .select('*')
+      .where('signing_key', validationData.validation_public_key)
+      .first()
+    const validationNetwork =
```
mvadari, line 71: Shouldn't this be the other way around?

src/connection-manager/connections.ts
```diff
@@ -55,11 +62,31 @@ async function handleWsMessageTypes(
...
+      validationData.networks ?? validationNetworkDb?.networks
```
mvadari, line 72: Shouldn't this be the other way around?

src/connection-manager/connections.ts
```diff
@@ -55,11 +62,31 @@ async function handleWsMessageTypes(
...
+    if (validationNetwork) {
+      validationData.ledger_fee = network_fee.get(validationNetwork)
+    }
```
mvadari, line 75: So is this setting the validator's default fee vote to the current network state?

src/connection-manager/agreement.ts
pdp2121: A validation would only have a `base_fee` field if it wants to change the current fee. Otherwise it would not be included, so we use the fee from the network (i.e retrieved from the ledger) to fill in ones that does not vote for a change

src/connection-manager/connections.ts
```diff
@@ -55,11 +62,31 @@ async function handleWsMessageTypes(
...
+    void agreement.handleValidation(validationData)
```
mvadari, line 76: Why is the type assertion not needed here anymore?

src/connection-manager/connections.ts
mvadari: Won't this override a fee vote by the validator? Shouldn't it only set this if `validationData.ledger_fee` isn't set?

pdp2121: the `ledger_fee` field is a field that I added (not included in the validation response from the stream) containing `base_fee`, `reserve_base`, `reserve_inc`. If we don't see one of these 3 fields in the validation (so no new fee is voted), then we will use `ledger_fee` to fill in the data.

src/connection-manager/connections.ts
pdp2121: validationData already has type assertion on line 61: `const validationData = data as ValidationRaw`

ckniffen (approved):

src/shared/types/index.ts
```diff
@@ -92,6 +98,21 @@ interface Agreement {
...
+interface Ballot {
+  signing_key: string
+  ledger_index: number
+  amendments?: string
+  base_fee?: number
+  reserve_base?: number
+  reserve_inc?: number
+}
+
+interface Fee {
```
mvadari, line 110: Can this be called `FeeVote` or something to be a bit more descriptive?

src/connection-manager/connections.ts
```diff
@@ -55,11 +62,35 @@ async function handleWsMessageTypes(
     if (ledger_hashes.includes(validationData.ledger_hash)) {
       validationData.networks = networks
     }
-    void agreement.handleValidation(data as ValidationRaw)
+
+    // Get network of the validation if ledger_hash is not in cache.
```
mvadari, line 66: This comment is confusing - there's no ledger_hash check

src/connection-manager/agreement.ts
mvadari, line 197: nit: should be camelCase

src/connection-manager/connections.ts
mvadari: NVM I see the line of code above.

mvadari: Seems like it's more the opposite - use the ledger hash data if it's not in the validators db.

src/connection-manager/connections.ts
```diff
@@ -55,11 +62,35 @@ async function handleWsMessageTypes(
...
+    const validationNetworkDb: DatabaseValidator | undefined = await query(
+      'validators',
+    )
+      .select('*')
+      .where('signing_key', validationData.validation_public_key)
+      .first()
+    const validationNetwork =
+      validationNetworkDb?.networks ?? validationData.networks
```
mvadari, line 74: nit: I think the code would be a bit easier to read if this was in a helper function.

src/connection-manager/connections.ts
```diff
@@ -16,6 +22,7 @@ const log = logger({ name: 'connections' })
 const ports = [443, 80, 6005, 6006, 51233, 51234]
 const protocols = ['wss://', 'ws://']
 const connections: Map<string, WebSocket> = new Map()
+const network_fee: Map<string, FeeVote> = new Map()
```
mvadari, line 25: nit: should also be camelCase

mvadari (approved):

PR #133, opened 2023-09-20
feat: add amendments enabled on each network

pdp2121:

Create a table with the following information to track amendments enabled on each network:

- AmendmentID

- NetworkID

- ledger_index

- tx_hash

- date

Also add a function to manually update the table using a JSON file (runs whenever re-deployed).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Works on staging.


ARCHITECTURE.md
```diff
@@ -108,6 +108,17 @@ This table keeps track of the manifests of the validators.
...
+This table keeps track of the amendments enabled on each network.
+
+| Key                  | Definition                                              |
+|----------------------|---------------------------------------------------------|
+| `amendment_id`       |The amendment id.                                        |
+| `networks`           |The network where the amendment has been enabled.        |
+| `ledger_index`       |The ledger where the amendment has been enabled.         |
+| `tx_hash`            |The transaction hash where the amendment has been enabled.|
+| `date`               |The date when the amendment has been enabled.            |
 ### `amendments_info`
```
mvadari, line 122:
```suggestion
| `date`               |The date when the amendment has been enabled.            |

### `amendments_info`
```

src/connection-manager/connections.ts
```diff
@@ -45,15 +56,79 @@ function subscribe(ws: WebSocket): void {
 }
 
 /**
- * Handles a WebSocket message received.
+ * Sends a ledger_entry WebSocket request to retrieve amendments enabled on a network.
+ *
+ * @param ws - A WebSocket object.
+ */
+function getAmendmentLedgerEntry(ws: WebSocket): void {
+  ws.send(
```
mvadari, line 64: Perhaps all of these `ws.send` functions could go in a different file, to satisfy the linter?

src/shared/database/setup.ts
```diff
@@ -205,6 +206,35 @@ async function setupNetworksTable(): Promise<void> {
...
+  if (!(await db().schema.hasColumn('amendments_enabled', 'date'))) {
+    await db().schema.alterTable('amendments_enabled', (table) => {
+      table.datetime('date').after('tx_hash')
+    })
+  }
+  if (!(await db().schema.hasColumn('amendments_enabled', 'ledger_index'))) {
+    await db().schema.alterTable('amendments_enabled', (table) => {
+      table.integer('ledger_index').after('networks')
+    })
+  }
```
mvadari, line 235: Was this just needed for testing? It won't be needed in prod, since the db doesn't exist yet.

src/shared/types/index.ts
```diff
@@ -29,6 +30,95 @@ interface StreamLedger {
...
+      ledger_hash: string
+      ledger_index: string
+      parent_close_time: number
+      parent_hash: string
+      seqNum: string
+      totalCoins: string
+      total_coins: string
+      transaction_hash: string
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Use any since we do not need to build Transaction model.
+      transactions: any[]
```
mvadari, line 84: Why not use types in xrpl.js?

src/shared/types/index.ts
```diff
@@ -29,6 +30,95 @@ interface StreamLedger {
...
+    Account: string
+    Amendment: string
+    Fee: string
+    LedgerSequence: string
+    Sequence: string
+    SigningPubKey: string
+    TransactionType: string
+    hash: string
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Use any since we do not need to build Transaction model.
+    meta: any
```
mvadari, line 105: ditto

src/shared/utils/index.ts
```diff
@@ -8,6 +8,7 @@ import logger from './logger'
 
 const log = logger({ name: 'utils' })
 const HTTPS_PORT = 51234
+const RIPPLE_EPOCH_DIFF = 0x386d4380
```
mvadari, line 11: This should be in decimal so it's easier to read (and matches the xrpl.js/Explorer code).

src/shared/data/amendments_enabled.json
```diff
@@ -0,0 +1,492 @@
...
+    },
+    {
+      "id": "2E2FB9CF8A44EB80F4694D38AADAE9B8B7ADAFD2F092E10068E61C98C4F092B0",
+      "ledger_index": 81986305,
+      "tx_hash": "EFE82B7155CE5B766AF343D98DAE6662C2713C99E760D610370D02338881B2F3",
+      "date": 745912751
+    }
+    ]
+  },
+  {"networks": "dev",
```
mvadari, line 312: This list will probably change given the reset now.

src/connection-manager/connections.ts
```diff
@@ -93,18 +168,80 @@ async function handleWsMessageTypes(
...
  * @param ws - A WebSocket object.
  * @param networks - The networks of the node we are trying to reach where it retrieves validations.
+ * @param entry - Whether source node is an entry node for the network.
  * @returns A Promise that resolves to void once a connection has been created or timeout has occured.
  */
 async function setHandlers(
   ip: string,
   ws: WebSocket,
   networks: string | undefined,
+  entry = false,
```
mvadari, line 244: `entry` has two meanings - an element or the place of entrance. It's a bit of a confusing name here (I had to look at the docstring to understand what it meant because I thought it was referring to the response at first) - perhaps something like `isInitialNode` would be a better name.

ARCHITECTURE.md
```diff
@@ -108,6 +108,17 @@ This table keeps track of the manifests of the validators.
...
+### `amendments_enabled`
+
+This table keeps track of the amendments enabled on each network.
+
+| Key                  | Definition                                              |
+|----------------------|---------------------------------------------------------|
+| `amendment_id`       |The amendment id.                                        |
+| `networks`           |The network where the amendment has been enabled.        |
+| `ledger_index`       |The ledger where the amendment has been enabled.         |
+| `tx_hash`            |The transaction hash where the amendment has been enabled.|
```
mvadari, line 120: This table should be aligned - you can just add spaces to align it all.

src/shared/types/index.ts
pdp2121: Currently, the `Transaction` type in xrpl.js is missing `EnableAmendment` tx type:
https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/models/transactions/transaction.ts#L90-L129
I will create a PR to update it and add a TODO to update the code here accordingly once the change has been released

src/connection-manager/utils.ts
```diff
@@ -0,0 +1,220 @@
...
+  TxEnableAmendmentResponse,
+  ValidationRaw,
+} from '../shared/types'
+import { rippleTimeToUnixTime } from '../shared/utils'
+
+import agreement from './agreement'
+import { handleManifest } from './manifests'
+
+const LEDGER_HASHES_SIZE = 10
+const AMENDMENT_LEDGER_ENTRY_INDEX =
```
ckniffen, line 25: This value is exported by the `xrpl` package.  Maybe we move `xrpl` into `dependencies`.

src/shared/utils/index.ts
```diff
@@ -8,6 +8,7 @@ import logger from './logger'
 
 const log = logger({ name: 'utils' })
 const HTTPS_PORT = 51234
+const RIPPLE_EPOCH_DIFF = 946684800
```
ckniffen, line 11: I think the `xrpl` has a utility method for this.

ARCHITECTURE.md
```diff
@@ -108,6 +108,18 @@ This table keeps track of the manifests of the validators.
...
+
+This table keeps track of the amendments enabled on each network.
+
+| Key                  | Definition                                               |
+|----------------------|----------------------------------------------------------|
+| `amendment_id`       |The amendment id.                                         |
+| `networks`           |The network where the amendment has been enabled.         |
+| `ledger_index`       |The ledger where the amendment has been enabled.          |
+| `tx_hash`            |The transaction hash where the amendment has been enabled.|
+| `date`               |The date when the amendment has been enabled.             |
```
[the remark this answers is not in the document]
pdp2121: It is of type `Date` but should show both date & time. How should I handle that?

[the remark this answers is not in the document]
ckniffen: I would just update the documentation line. `The date and time`

ARCHITECTURE.md
```diff
@@ -108,6 +108,18 @@ This table keeps track of the manifests of the validators.
...
+
+This table keeps track of the amendments enabled on each network.
+
+| Key                  | Definition                                               |
+|----------------------|----------------------------------------------------------|
+| `amendment_id`       |The amendment id.                                         |
+| `networks`           |The network where the amendment has been enabled.         |
+| `ledger_index`       |The ledger where the amendment has been enabled.          |
+| `tx_hash`            |The transaction hash where the amendment has been enabled.|
+| `date`               |The date and time when the amendment has been enabled.    |
```
JST5000, line 121: Is this in ripple epoch time or POSIX time?

src/connection-manager/connections.ts
```diff
@@ -26,89 +32,22 @@ const network_fee: Map<string, Fee> = new Map()
...
-  }
-}
-
 /**
  * Sets the handlers for each WebSocket object.
  *
  * @param ip - The ip address of the node we are trying to reach.
  * @param ws - A WebSocket object.
  * @param networks - The networks of the node we are trying to reach where it retrieves validations.
+ * @param isInitialNode - Whether source node is an entry/initial node for the network.
```
JST5000, line 43: I'm not sure what an entry node is in this case - is it an XRPL specific term, or is it just the first node you've encountered when crawling? (If there's a docs page which is relevant, can you link to it?)

JST5000: From reading the code more, it seems this field is used to look up manually added values - in which case I would rename this to be more "functional" - meaning something like `lookUpAmendmentsManually`. That way it's easier to process this single function's purpose on it's own without knowing the architecture of the project (e.g. that we start by looking up nodes with amendment data we pre-defined)

src/connection-manager/update-amendments-from-json.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Save amendment enabled on a network to the database.
+ *
+ * @param enabledAmendment - The input amendment.
+ *
+ * @returns Void.
+ */
+async function saveAmendmentsEnabled(
+  enabledAmendment: AmendmentEnabled,
+): Promise<void> {
+  await query('amendments_enabled')
```
JST5000, line 34: Since `addAmendmentsDataFromJSON` has a `log.info` explaining what the function is doing, should this log as well?

src/connection-manager/update-amendments-from-json.ts
```diff
@@ -0,0 +1,63 @@
...
+        amendment_id: amendment.id,
+        networks: networkData.networks,
+        ledger_index: amendment.ledger_index,
+        tx_hash: amendment.tx_hash,
+        date: new Date(rippleTimeToUnixTime(amendment.date)),
+      }
+      await saveAmendmentsEnabled(enabledData)
+    })
+  })
+  log.info('Finished Enabled Amendment Data from JSON File.')
```
JST5000, line 62: nit: 
```suggestion
  log.info('Finished adding Enabled Amendment Data from JSON File.')
```

src/connection-manager/utils.ts
```diff
@@ -0,0 +1,219 @@
...
+      expand: true,
+    }),
+  )
+}
+
+/**
+ * Sends a tx WebSocket request to retrieve EnableAmendment transaction details.
+ *
+ * @param ws - A WebSocket object.
+ * @param transaction -- The hash of the transaction..
```
JST5000, line 78: nit:
```suggestion
 * @param transaction -- The hash of the transaction.
```

ARCHITECTURE.md
pdp2121: It's converted to js `Date` type instead of a number so I'm not sure it needs clarification here (it will show YYYY-MM-DD HH:MM:SS in the database).

src/connection-manager/update-amendments-from-json.ts
pdp2121: `addAmendmentsDataFromJSON` only execute once in a while, while `saveAmendmentsEnabled` will log for each amendments and will swamp the log so I don't think it's necessary here (otherwise we will need to log whenever we add/update one row to the db for other updates as well).

src/connection-manager/connections.ts
pdp2121: For this part I want to look for fee data to apply the fee for all validators in a network that doesn't vote for a new fee, and I want to do it only once per network. That's why I added a boolean to check if the node is the initial node that we added manually at first for each network in order to start crawling, so the the ledgerEntry calls only execute once per network per ledger. Sorry for it being confusing for you

ARCHITECTURE.md
JST5000: I think it's still important to mention because that changes how you interpret the number. (The `Date` type is how it's formatted, but knowing if it's ripple epoch changes what that date represents. If you directly change a ripple epoch number into a Date, it'll give you the wrong date)

src/connection-manager/connections.ts
JST5000: I see, that makes sense :) 

Can you add an in-line comment above `if (isInitialNode) {` explaining that?

My thought process: In general, I tend to think if there's a question a code reviewer has that seems reasonable, but the code is working as is the resolution is to make that same answer more discoverable in the code. 

The best way to do that is to simplify the logic. Here, since you're only using this variable once, I checked if it was reasonable to pull that parameter / action out of this function so it can be called once separately. In this case I don't think that makes sense.

Then since the logic seems fine, a comment where the confusion arose seems like a reasonable answer.

src/connection-manager/update-amendments-from-json.ts
JST5000: Makes sense :)

JST5000 (approved):
LGTM

ARCHITECTURE.md
mvadari: It'd probably be easier to store it in ripple epoch, tbh. You can convert when outputting.

src/connection-manager/connections.ts
```diff
@@ -1,22 +1,24 @@
...
-import agreement from './agreement'
-import { handleManifest } from './manifests'
+import {
+  getAmendmentLedgerEntry,
+  handleWsMessageLedgerEnableAmendments,
+  handleWsMessageLedgerEntryAmendments,
+  handleWsMessageSubscribeTypes,
+  handleWsMessageTxEnableAmendments,
+  subscribe,
+} from './utils'
```
mvadari, line 21: IMO this file should be called something like `wsHandling` (probably a better name is needed) because it's storing all the websocket handling. `utils` is too vague.

src/connection-manager/connections.ts
```diff
@@ -120,6 +55,15 @@ async function setHandlers(
       void saveNodeWsUrl(ws.url, true)
       connections.set(ip, ws)
       subscribe(ws)
+
+      // Use LedgerEntry to look for fee data to apply the fee for all validators
+      // in a network that doesn't vote for a new fee. We want to do it only once
+      // per network. This part will check if the node is the initial node that we
```
mvadari, line 61: Why do we want to do this only once per network?

mvadari: And why not make that a separate call that happens?

pdp2121: Other details for these already added amendments will also be added manually via the `amendments_enabled.json` file (but only ones that had been enabled via EnableAmendment tx in the past, those are added on genesis will have `ledger_index`, `tx_hash` and `date` fields blank

src/connection-manager/update-amendments-from-json.ts
```diff
@@ -0,0 +1,63 @@
...
+const filePath = 'src/shared/data/amendments_enabled.json'
+
+interface AmendmentEnabledJson {
+  networks: string
+  amendments: Array<{
+    id: string
+    networks: string
+    ledger_index: number
+    tx_hash: string
+    date: number
```
ckniffen, line 20: The date should be different on every network.  This data model makes it seem like the date is shared.

src/connection-manager/update-amendments-from-json.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import { query } from '../shared/database/utils'
+import { AmendmentEnabled } from '../shared/types'
+import logger from '../shared/utils/logger'
+
+const log = logger({ name: 'database-agreement' })
+const filePath = 'src/shared/data/amendments_enabled.json'
+
+interface AmendmentEnabledJson {
+  networks: string
```
ckniffen, line 14: It should not be called networks because each are different.

src/shared/data/amendments_enabled.json
```diff
@@ -0,0 +1,386 @@
...
+    },
+    {
+      "id": "F1ED6B4A411D8B872E65B9DCB4C8B100375B0DD3D62D07192E011D6D7F339013",
+      "ledger_index": 38410497,
+      "tx_hash": "E2B96A64B0C77439AFF495837D2E560D508404EBC634BC7F0A56BABB607D6F38",
+      "date": 739312992
+    }
+  ]
+  },
+  {"networks":"hooks-test",
```
ckniffen, line 370: devnet appears to be missing.

src/connection-manager/update-amendments-from-json.ts
```diff
@@ -0,0 +1,63 @@
...
+  amendments: Array<{
+    id: string
+    networks: string
```
ckniffen, line 17: in the json the `networks` field is not populated.

src/connection-manager/update-amendments-from-json.ts
pdp2121: It is called `networks` across other tables as well even though all should be a 1-1 match. I have created an issue to fix them:
https://github.com/ripple/validator-history-service/issues/165

ckniffen (approved):

PR #136, opened 2023-09-20
feat: add amendments to validators endpoint

pdp2121:

Add voting data to validators related endpoints.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


src/api/routes/v1/validator.ts
```diff
@@ -42,6 +42,10 @@ interface ValidatorResponse {
   } | null
   partial: boolean
   unl: boolean
+  amendments?: string | Array<{ id: string; name: string }>
```
ckniffen, line 45: You should remove `string` being allowed.  I dont think there is much use for it to be a string.

jonathanlei (approved):

src/api/routes/v1/validator.ts
```diff
@@ -42,6 +42,10 @@ interface ValidatorResponse {
   } | null
   partial: boolean
   unl: boolean
+  amendments: Array<{ id: string; name: string }> | null
```
mvadari, line 45: When is this `null`?

src/api/routes/v1/utils.ts
```diff
@@ -26,6 +26,27 @@ export function formatAgreementScore(agreement: AgreementScore): {
...
+export async function formatAmendments(
+  amendmentsDb: string,
+): Promise<Array<{ id: string; name: string }>> {
+  const res: Array<{ id: string; name: string }> = []
+  const amendmentsList = amendmentsDb.split(',')
+  amendmentsList.forEach(async (amendment) => {
+    const info = await query('amendments_info')
+      .select('id', 'name')
+      .where('id', amendment)
+      .first()
```
mvadari, line 44: Why return both the ID and the name?

ckniffen: One is the pretty name the other is the id.

mvadari: Yes, but why are both needed here?

pdp2121: It would be for the API for validators to show the list of both amendment_id and name. Also, it would be for the Voting Tab on the explorer for validator as you would not need to search for name by id there, just need to make one API call. In VHS db ballot, each validator has a field amendments which is a string of ids comma separated, which is why this function is needed.

ckniffen (approved):

PR #169, opened 2023-11-21
feat: Add enable amendment eta

pdp2121:

Add amendment eta time after majority gained tracking by `amendments_incoming` table.

The WebSocket handling process for amendments is now as below:

- When ws connection open, subscribe and run ledger_entry once per network via initial node to get amendments already enabled on the network at the time connection starts.
- When listening to subscribed ledger stream, when ledger index is flag + 1, send a ledger request with `transactions` and `expand` options set to `true` to scan for `EnableAmendment` transactions (only processed once per network via initial node).
- If `EnableAmendment` transaction found:

1. if no Flags, a tx request is sent to retrieve more details about the amendment, a new entry will be added to `amendments_status` table, with `date` field added to reflect the release date, and `eta` removed as the amendment has already been enabled.
2. if Majority Gain Flags, a new `eta` data will be added to `amendments_status` table.
3. if Majority Lost Flags, delete the entry from `amendments_status` table.

- Ballots will be received via validation stream on flag - 1.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


ARCHITECTURE.md
```diff
@@ -117,12 +117,22 @@ This table keeps track of the amendments enabled on each network.
...
 |----------------------|----------------------------------------------------------|
 | `amendment_id`       |The amendment id.                                         |
-| `networks`           |The network(s) where the amendment has been enabled.         |
+| `networks`           |The network(s) where the amendment has been enabled.      |
 | `ledger_index`       |The ledger where the amendment has been enabled.          |
 | `tx_hash`            |The transaction hash where the amendment has been enabled.|
 | `date`               |The date and time when the amendment has been enabled.    |
 
 
+### `amendments_incoming`
```
mvadari, line 126: Can we merge `amendments_enabled` and `amendments_incoming`? They seem to contain the same unique index and have similar data.

pdp2121: Should we rename it or keep it as `amendments_enabled`?

pdp2121: I was thinking the same thing but just struggle to figure out a name

mvadari: Could be `amendment_networks` or `amendment_status` or something

ckniffen (approved):

src/shared/database/setup.ts
```diff
@@ -212,10 +212,10 @@ async function setupNetworksTable(): Promise<void> {
   })
 }
 
-async function setupAmendmentsEnabledTable(): Promise<void> {
-  const hasAmendmentsEnabled = await db().schema.hasTable('amendments_enabled')
-  if (!hasAmendmentsEnabled) {
-    await db().schema.createTable('amendments_enabled', (table) => {
+async function setupAmendmentsStatusTable(): Promise<void> {
```
mvadari, line 215: Doesn't this table need an `eta` element?

pdp2121: I was thinking of using just the date before and compare with `Date.now()` but you are right since there should be a gap between eta and actual flag+1 time. I will work on adding that back.

src/api/routes/v1/amendments.ts
```diff
@@ -222,11 +229,12 @@ async function calculateConsensus(
...
 
 /**
- * Retrieves amendments enabled on a network.
+ * Retrieves amendments in voting on a network.
  *
  * @param id - The network id.
- * @returns List of enabled amendments.
+ * @returns List of amendments in voting.
  */
+// eslint-disable-next-line max-lines-per-function, max-statements -- Disabled for this function.
```
JST5000, line 237: nit: This comment is describing what the eslint-disable is doing, but should describe why it's necessary. (In this case, explaining why this function needs to be so long instead of being broken up)

src/api/routes/v1/amendments.ts
```diff
@@ -54,7 +55,7 @@ interface AmendmentInVotingMap {
   }
 }
 
-interface AmendmentsEnabled extends AmendmentsInfo {
+interface AmendmentStatusExtended extends AmendmentsInfo {
```
JST5000, line 58: Based on the name, I would expect `AmendmentStatusExtended` to extend `AmendmentStatus`

(Also, if these are just optional parameters, would it make sense to include them as part of the original object as optional fields instead of having a separate type?)

src/shared/database/update-amendments-from-json.ts
```diff
@@ -1,56 +1,41 @@
...
-  enabledAmendment: AmendmentEnabled,
-): Promise<void> {
-  await query('amendments_enabled')
-    .insert(enabledAmendment)
+  await query('amendments_status')
+    .insert(statusAmendment)
     .onConflict(['amendment_id', 'networks'])
     .merge()
-    .catch((err) => log.error('Error Saving Enabled Amendment', err))
+    .catch((err) => log.error('Error Saving Status Amendment', err))
```
JST5000, line 38: nit:
```suggestion
    .catch((err) => log.error('Error Saving Amendment Status', err))
```

src/shared/types/index.ts
```diff
@@ -56,12 +57,13 @@ interface LedgerResponseCorrected extends LedgerResponse {
   >
 }
 
-interface AmendmentEnabled {
+interface AmendmentStatus {
```
JST5000, line 60: A docstring explaining what this represents would be helpful. 

Previously this just represented amendments that were enabled on a network.

Now it's representing two things:
1. An enabled amendment
2. An amendment which is being voted in. 

A different way you could represent that which might be more accurate would be 

```

interface AmendmentEnabled {
  amendment_id: string
  networks?: string
  ledger_index?: number
  tx_hash: string
  date?: Date
}

interface AmendmentPending extends AmendmentEnabled {
  tx_hash?: string
  eta?: Date
}

type AmendmentStatus = AmendmentEnabled | AmendmentPending
```

ckniffen: `tx_hash` should be optional on `AmendmentEnabled` too as we might not know when it was enabled on some networks such as if history was lost or it was activated at genesis.

JST5000 (approved):
LGTM

PR #195, opened 2024-01-22
fix: validator report timestamp

pdp2121:

There was a bug in response format for individual validator reports (`/validator/{pubkey}/reports`) that caused the date to be in the future for timezone later than UTC. Set the hour of the date to be 0 instead of 23 would fix this issue, and also correspond with how daily scores save their dates and the `/validator_reports`.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

The effect of this change has been tested locally by changing browser's timezone

## Future Tasks

Date time on the Explorer should all be in UTC with a specified note to ensure consistency in the data that users in various regions see.

### DIFF
--- a/src/api/routes/v1/validator-report.ts
+++ b/src/api/routes/v1/validator-report.ts
@@ -35,7 +35,7 @@ function formatResponse(response: DatabaseResponse): ScoreResponse {
   } = response
   const denominator = validated + missed
   const score: number = denominator === 0 ? 0 : validated / denominator
-  date.setHours(23, 0, 0, 0)
+  date.setHours(0, 0, 0, 0)
 
   return {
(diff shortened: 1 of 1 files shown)


ckniffen (approved):

mvadari:
Will this work okay when it changes and isn't starting from scratch?

pdp2121:
> Will this work okay when it changes and isn't starting from scratch?

@mvadari The dates in the database are already zeroed out. This change does not affect data in db, just on api side when it convert the date so I don't think it would have any significant effect when we deploy halfway (at least on VHS and Explorer, not sure about other apps using this api)

mvadari (approved):

PR #203, opened 2024-02-01
Fix EnableAmendment tx scanning process and add tests

pdp2121:

The amendment's ETA when reached majority or enabled was not captured by the VHS at the moment. This PR attempts to fix the issue by:

- When an `EnableAmendment` transaction is found, use ledger's `close_time` instead of transaction's date field in the `ledger` response, which does not exist (This will also simplify the code as extra `tx` request does not need to be sent to capture the date).
- Remove `date` from ledger response types transaction.
- Add test for `handleWsMessageLedgerEnableAmendments` function which scans for `EnableAmendment` transaction at Flag + 1 ledger.
- Add more logs to the amendment scanning process.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Leave on staging for couple days to ensure all flag+1 ledgers are scanned.


mvadari (commented):
Can the testcases be made smaller so the files aren't so massive? Perhaps most of the unused transactions can be removed.

test/connections/amendment.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  beforeAll(async () => {
+    await setupTables()
+  })
+
+  afterAll(async () => {
+    await destroy()
+  })
+
+  beforeEach(async () => {
+    await query('amendments_status').delete('*')
```
JST5000, line 20: Why are you deleting the amendments_status before and after each test?

test/connections/amendment.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    expect(amendmentStatus[0].networks).toBe('main')
+    expect(amendmentStatus[0].ledger_index).toBe(84206081)
+    expect(amendmentStatus[0].tx_hash).toBe(
+      'CA4562711E4679FE9317DD767871E90A404C7A8B84FAFD35EC2CF0231F1F6DAF',
+    )
+    expect(amendmentStatus[0].date.toISOString()).toBe(
+      '2023-11-27T14:44:30.000Z',
+    )
+    expect(amendmentStatus[0].eta).toBe(null)
+  })
```
JST5000, line 70: Should there be a test for a negative case? (e.g. someone passes in a ledger object which isn't "corrected" or isn't a flag ledger?)

src/shared/types/index.ts
```diff
@@ -43,7 +43,6 @@ interface LedgerCorrected extends LedgerXRPL {
     (Transaction | PseudoTransaction) & {
       metaData?: TransactionMetadata
       hash: string
-      date: number
```
mvadari, line 46: Why was this removed?

test/connections/amendment.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    await query('amendments_status').delete('*')
+  })
+
+  test('Correctly finds EnableAmendment tx with tfGotMajority Flag (eta available) from ledger response', async () => {
+    await handleWsMessageLedgerEnableAmendments(
+      ledgerResponseGotMajority as LedgerResponseCorrected,
+      'main',
+    )
+
+    await flushPromises()
```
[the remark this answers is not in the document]
mvadari: It makes sure all the awaited functions have fully run (just to prevent race conditions).

mvadari (approved):

test/connections/amendment.test.ts
pdp2121: It's to clean up the amendments_status table in the db between each test otherwise the next test might have data from the previous one

src/shared/types/index.ts
pdp2121: This was added by mistake as I thought there's a date field in the expanded ledger response but actually there isn't and we can use the `close_time` instead

JST5000 (approved):
LGTM

PR #208, opened 2024-02-08
fix: Reconnect immediately to established connections

pdp2121:

The VHS connections is experiencing lost connections to a large number of nodes (only 40-50 node connections remained in February, with a low of 9-15 connections in January). This is because those node connections failed after being connected and the VHS does not try to reconnect until the next hour (bulk connection also led to slow response and high resource consumption). The frequent closing codes seen so far include:

- **1008**: Policy error: client is too slow. (Most frequent)
- **1006**: Abnormal Closure: The connection was closed abruptly without a proper handshake or a clean closure.
- **1005**: No Status Received: An empty or undefined status code is used to indicate no further details about the closure.

This PR will allow the VHS to reconnect to established connections after failure, instead of waiting for an hour for a bulk connections attempt. This would also help fixing the problem in https://github.com/ripple/validator-history-service/pull/203.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Test on staging:
- Monitor the number of messages for failed connection before and after connection
- Extra: merge with https://github.com/ripple/validator-history-service/pull/203 and monitor the frequency of flag + 1 ledgers scanned by entry nodes on each network.


mvadari (approved):

src/connection-manager/connections.ts
```diff
@@ -28,6 +28,8 @@ const networkFee: Map<string, FeeVote> = new Map()
 const CM_INTERVAL = 60 * 60 * 1000
 const WS_TIMEOUT = 10000
 const REPORTING_INTERVAL = 15 * 60 * 1000
+const CLOSING_CODES = [1005, 1006, 1008]
```
JST5000, line 31: Please add a comment similar to what you included in the PR description for what these closing codes mean / why they mean we should reconnect (as opposed to just accepting that it should be closed)

src/connection-manager/connections.ts
```diff
@@ -97,7 +99,28 @@ async function setHandlers(
...
+          )
+          // Open a new Websocket connection for the same url
+          const newWS = new WebSocket(ws.url, { handshakeTimeout: WS_TIMEOUT })
+          // Clean up the old Websocket connection
+          connections.delete(ip)
+          ws.terminate()
+          resolve()
+
+          await setHandlers(ip, newWS, networks, isInitialNode)
+          return
```
JST5000, line 121: Can you add a comment explaining why you return here instead of allowing the rest of the logic to run?

src/connection-manager/connections.ts
```diff
@@ -97,7 +99,28 @@ async function setHandlers(
         )
       }
     })
-    ws.on('close', () => {
+    ws.on('close', async (code, reason) => {
+      if (connectionsInitialized) {
+        log.error(
+          `Websocket closed for ${ws.url} on ${
+            networks ?? 'unknown network'
```
mvadari, line 106: `networks ?? 'unknown network'` should be a variable for ease of use/readability

src/connection-manager/connections.ts
pdp2121: The return statement comes inside a nested if statement so I think it suffice to just add a comment.

JST5000 (approved):
LGTM

PR #209, opened 2024-02-22
fix: connections reconnect after inactivity

pdp2121:

On the VHS, even when a WebSocket connection is open, it stops receiving subscription messages for some periods and resumes. The issue has been raised here to the Core Ledger:
https://github.com/XRPLF/rippled/issues/3180
It happened more often for some nodes than others. One approach to fix it in the immediate term is to reconnect after the ws remains open but no message has arrived in 1 minute.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


pdp2121:
This is outdated

PR #217, opened 2024-03-06
fix: `validators` api call mixed networks

pdp2121:

Currently the `validators/{networks}` call still uses chain matching. This led to test validators get pulled to main when there's a test validator validates on main accidentally, and vice versa. 

for example: if after a leakage, a test validator has chain `main` and networks `test`, all of the mainnet validators will show up on `validators/test`, which is incorrect.

We should just use `networks` field for `validators/{networks}` itself to mitigate this effect, and also, `networks` is updated more frequently using `ledger_hash`, which makes it more accurate compared to chain calculation.

In the future, we should revise chain calculation and agreement score calculation since there's some minor issues there (will create a ticket for it). The inclusion of `network-id` on the validation stream would also be a permanent solution, but it is still WIP.

### Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation Updates
- [ ] Release

## Test Plan

Leakage should be fixed in staging

### DIFF
--- a/src/api/routes/v1/utils.ts
+++ b/src/api/routes/v1/utils.ts
@@ -53,8 +53,8 @@ export async function formatAmendments(
 /**
  * Get param type from validator API call.
  *
- * @param param - The UNL/Networks of the chain.
- *
+ * @param param - The input parameter.
+ * @returns The type of the input parameter (unl/networks/unknown).
  */
 export async function getParamType(
   param: string | undefined,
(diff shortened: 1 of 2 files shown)


justinr1234 (approved):

ckniffen:
Is chain calculation based on ledger index alone?

pdp2121:
> Is chain calculation based on ledger index alone?

For the most part, yes, if you look here:
https://github.com/ripple/validator-history-service/blob/main/src/connection-manager/chains.ts#L193-L245

Meanwhile, `networks` uses recent `ledger_hash` from `ledger` stream, which is the desired behavior as these ledgers have been closed.

JST5000 (approved):

PR #223, opened 2024-03-28
feat: add amendments status backtracking

pdp2121:

Currently we are using `p2p.livenet.ripple.com` to retrieve amendments enablement via ledger + 1 for mainnet. However, its connection is very inconsistent, led to missing information.

We should backtrack every 30 minutes using clio for mainnet to retrieve this information in case the connections get lost.  We would also backtrack major networks like testnet and devnet to ensure we have adequate data for these networks.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Ran without breaking on staging

### DIFF
--- a/src/connection-manager/connections.ts
+++ b/src/connection-manager/connections.ts
@@ -13,6 +13,7 @@ import { FeeVote, LedgerResponseCorrected } from '../shared/types'
 import logger from '../shared/utils/logger'
 
 import {
+  backtrackAmendmentStatus,
   getAmendmentLedgerEntry,
   handleWsMessageLedgerEnableAmendments,
   handleWsMessageLedgerEntryAmendments,
@@ -28,6 +29,7 @@ const networkFee: Map<string, FeeVote> = new Map()
 const CM_INTERVAL = 60 * 60 * 1000
 const WS_TIMEOUT = 10000
 const REPORTING_INTERVAL = 15 * 60 * 1000
+const BACKTRACK_INTERVAL = 30 * 60 * 1000
 
 // The frequent closing codes seen so far after connections established include:
 //  1008: Policy error: client is too slow. (Most frequent)
@@ -243,8 +245,14 @@ export default async function startConnections(): Promise<void> {
     await fetchAmendmentInfo()
     await clearConnectionsDb()
     await createConnections()
+    await backtrackAmendmentStatus()
     setInterval(() => {
+      void fetchAmendmentInfo()
       void createConnections()
     }, CM_INTERVAL)
+
+    setInterval(() => {
+      void backtrackAmendmentStatus()
+    }, BACKTRACK_INTERVAL)
   }
 }
--- a/src/connection-manager/wsHandling.ts
+++ b/src/connection-manager/wsHandling.ts
@@ -1,6 +1,11 @@
 /* eslint-disable import/max-dependencies -- Disbale for this file which uses a lot of types. */
 import WebSocket from 'ws'
-import { LedgerEntryResponse, rippleTimeToUnixTime } from 'xrpl'
+import {
+  Client,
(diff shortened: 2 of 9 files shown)


src/connection-manager/wsHandling.ts
```diff
@@ -225,3 +235,69 @@ export async function handleWsMessageLedgerEnableAmendments(
...
+
+/**
+ * Backtracking a network to update amendment status in case of Websocket disconnection.
+ *
+ * @param networks - The network being tracked.
+ * @param url - The Faucet URL of the network.
+ * @returns Void.
+ */
+async function backtrackNetworkAmendmentStatus(
+  networks: string,
```
ckniffen, line 247: Make this param singular.

ckniffen (approved):

src/connection-manager/connections.ts
```diff
@@ -243,8 +245,14 @@ export default async function startConnections(): Promise<void> {
...
     await clearConnectionsDb()
     await createConnections()
+    await backtrackAmendmentStatus()
     setInterval(() => {
+      void fetchAmendmentInfo()
       void createConnections()
     }, CM_INTERVAL)
+
+    setInterval(() => {
+      void backtrackAmendmentStatus()
```
justinr1234, line 255: Technically speaking you need to have the ability to cancel a previous request in the event this request isn’t finished by the next interval tick (even if unlikely, it could happen)

src/connection-manager/wsHandling.ts
```diff
@@ -225,3 +235,69 @@ export async function handleWsMessageLedgerEnableAmendments(
...
+ * @param network - The network being tracked.
+ * @param url - The Faucet URL of the network.
+ * @returns Void.
+ */
+async function backtrackNetworkAmendmentStatus(
+  network: string,
+  url: string,
+): Promise<void> {
+  try {
+    log.info(`Backtracking to update amendment status for ${network}...`)
```
justinr1234, line 251: Are the faucet url to debug

src/connection-manager/wsHandling.ts
```diff
@@ -225,3 +235,69 @@ export async function handleWsMessageLedgerEnableAmendments(
...
+    ) {
+      const ledger: LedgerResponse = await client.request({
+        command: 'ledger',
+        transactions: true,
+        ledger_index: index,
+        expand: true,
+      })
+
+      await handleWsMessageLedgerEnableAmendments(
+        ledger as LedgerResponseCorrected,
```
justinr1234, line 277: Can you avoid `as`?

src/connection-manager/wsHandling.ts
```diff
@@ -225,3 +235,69 @@ export async function handleWsMessageLedgerEnableAmendments(
...
+        network,
+      )
+    }
+
+    await client.disconnect()
+
+    log.info(`Finished backtracked amendment status for ${network}...`)
+  } catch (error) {
+    log.error(
+      `Failed to backtrack amendment status for ${network} due to error: ${String(
```
justinr1234, line 287: It shouldn’t be necessary to wrap in String since toString is automatically called anyways

justinr1234 (approved):
Left optional comments and approved

src/connection-manager/wsHandling.ts
pdp2121: There's a missing field `hash` for the `BaseTransaction` in xrpl.js (which should be fixed), so I would need to cast to `LedgerResponseCorrected` which I created to capture the hash. Once this is fixed in our client library then there would be no need for custom field or casting.

src/connection-manager/wsHandling.ts
pdp2121: This is to avoid linting issue

Issue #278, opened 2024-10-18
Use `feature` RPC to fetch amendments name


pdp2121:
Similar to https://github.com/ripple/explorer/issues/1001

PR #291, opened 2024-12-19
feat: Use feature for amendment name

pdp2121:

Resolve: https://github.com/ripple/validator-history-service/issues/278

### Context of Change

Missing new amendments due to the change in rippled code structure to parse name:

Screenshot 2024-12-19 at 11 32 43 AM

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Data should show up:

Screenshot 2024-12-19 at 11 58 52 AM

VHS is lacking tests at the moment. There will be a separate ticket to add comprehensive tests throughout the code

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
   "homepage": "https://github.com/ripple/validator-history-service#readme",
   "devDependencies": {
     "@types/axios": "^0.14.0",
-    "@types/create-hash": "^1.2.2",
     "@types/express": "4.17.21",
     "@types/jest": "^26.0.19",
     "@types/nconf": "^0.10.0",
@@ -62,7 +61,6 @@
     "@types/bunyan": "^1.8.7",
     "axios": "^0.21.1",
     "bunyan": "^1.8.15",
-    "create-hash": "^1.2.0",
     "dotenv": "^16.3.1",
     "express": "4.21.2",
     "knex": "2.5.1",
--- a/src/shared/database/amendments.ts
+++ b/src/shared/database/amendments.ts
@@ -1,5 +1,6 @@
 import axios from 'axios'
 import createHash from 'create-hash'
+import { Client, FeatureResponse } from 'xrpl'
 
 import { AmendmentInfo } from '../types'
 import logger from '../utils/logger'
@@ -18,6 +19,12 @@ const RETIRED_AMENDMENT_REGEX =
 
 const AMENDMENT_VERSION_REGEX =
   /\| \[(?<amendmentName>[a-zA-Z0-9_]+)\][^\n]+\| (?<version>v[0-9]*\.[0-9]*\.[0-9]*|TBD) *\|/u
+
+export const NETWORKS_HOSTS = new Map([
+  ['main', 'ws://s2.ripple.com:51233'],
+  ['test', 'wss://s.altnet.rippletest.net:51233'],
+  ['dev', 'wss://s.devnet.rippletest.net:51233'],
+])
 
 // TODO: Clean this up when this PR is merged:
 // https://github.com/XRPLF/rippled/pull/4781
@@ -81,6 +88,35 @@ async function nameOfAmendmentID(): Promise<void> {
         deprecated,
       })
     })
+  }
+}
+
+/**
+ * @param network
+ * @param url
+ */
+async function getNetworkAmendmentName(
+  network: string,
+  url: string,
+): Promise<void> {
+  try {
+    log.info(`Backtracking to update amendment status for ${network}...`)
+    const client = new Client(url)
+    await client.connect()
+    const featureResponse: FeatureResponse = await client.request({
+      command: 'ledger',
+      ledger_index: 'validated',
+    })
+
+    await client.disconnect()
+
+    log.info(`Finished backtracked amendment status for ${network}...`)
+  } catch (error) {
+    log.error(
+      `Failed to backtrack amendment status for ${network} due to error: ${String(
+        error,
+      )}`,
+    )
   }
 }
 
--- a/src/shared/database/amendments.ts
+++ b/src/shared/database/amendments.ts
@@ -14,6 +14,8 @@ const log = logger({ name: 'amendments' })
 
 const amendmentIDs = new Map<string, { name: string; deprecated: boolean }>()
 const rippledVersions = new Map<string, string>()
+// TODO: Use feature RPC instead when this issue is fixed and released:
+// https://github.com/XRPLF/rippled/issues/4730
 const RETIRED_AMENDMENTS = [
   'MultiSign',
   'TrustSetAuth',
--- a/src/shared/database/amendments.ts
+++ b/src/shared/database/amendments.ts
@@ -96,14 +96,14 @@ async function fetchNetworkAmendments(
 }
 
 /**
- * Fetch an amendment info from a network.
+ * Find an amendment info from a network and add to current map.
  *
  * @param amendment_id - The id of the amendment to fetch.
  * @param url - The Faucet URL of the network.
  *
  * @returns True if the amendment is found, false otherwise.
  */
-async function fetchSingleAmendment(
+async function findSingleAmendment(
   amendment_id: string,
   url: string,
 ): Promise<boolean> {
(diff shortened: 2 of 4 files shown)


src/shared/database/amendments.ts
```diff
@@ -10,77 +14,163 @@ const log = logger({ name: 'amendments' })
...
+  'fix1523',
+  'fix1528',
+]
 
 const AMENDMENT_VERSION_REGEX =
   /\| \[(?<amendmentName>[a-zA-Z0-9_]+)\][^\n]+\| (?<version>v[0-9]*\.[0-9]*\.[0-9]*|TBD) *\|/u
 
-// TODO: Clean this up when this PR is merged:
-// https://github.com/XRPLF/rippled/pull/4781
+export const NETWORKS_HOSTS = new Map([
```
ckniffen, line 40: Should this list come from the database.  That way more networks would be supported.

package.json
```diff
@@ -31,7 +31,6 @@
   "homepage": "https://github.com/ripple/validator-history-service#readme",
   "devDependencies": {
     "@types/axios": "^0.14.0",
-    "@types/create-hash": "^1.2.2",
```
ckniffen, line 34: why didnt package-lock change with this?

src/shared/database/amendments.ts
```diff
@@ -1,5 +1,9 @@
 import axios from 'axios'
-import createHash from 'create-hash'
+import { Client, ErrorResponse } from 'xrpl'
+import {
+  FeatureAllResponse,
+  FeatureOneResponse,
+} from 'xrpl/dist/npm/models/methods/feature'
```
ckniffen, line 6: Are these not exported by `xrpl` normally?

package.json
ckniffen: Turns out create-hash is still used by `xrpl-validator-domains`.  That library needs love.  It is using xrpl@3 still.

src/shared/database/amendments.ts
ckniffen: nvm i just realized this moved.

src/shared/database/amendments.ts
```diff
@@ -10,77 +14,163 @@ const log = logger({ name: 'amendments' })
...
+      await client.request({
+        command: 'feature',
+        feature: amendment_id,
+      })
+
+    if ('result' in featureResponse) {
+      const feature = featureResponse.result[amendment_id]
+      amendmentIDs.set(amendment_id, {
+        name: feature.name,
+        deprecated: RETIRED_AMENDMENTS.includes(feature.name),
```
ckniffen, line 124: I wonder if deprecated is the right term.  There are some that are fully part of the codebase now in the retired list.

src/shared/database/amendments.ts
```diff
@@ -10,77 +14,163 @@ const log = logger({ name: 'amendments' })
...
+    const featureResponse: FeatureOneResponse | ErrorResponse =
+      await client.request({
+        command: 'feature',
+        feature: amendment_id,
+      })
+
+    if ('result' in featureResponse) {
+      const feature = featureResponse.result[amendment_id]
+      amendmentIDs.set(amendment_id, {
+        name: feature.name,
```
ckniffen, line 123: Is there more that should be stored?

src/shared/database/amendments.ts
pdp2121: No it is not being exported in index file. We'll need to fix that in xrpl.js. I'll add a TODO

src/shared/database/amendments.ts
```diff
@@ -10,77 +14,163 @@ const log = logger({ name: 'amendments' })
...
-  // The Amendment ID is the hash of the Amendment name
-  const amendmentNames = await fetchAmendmentNames()
-  if (amendmentNames !== null) {
-    amendmentNames.forEach((deprecated, name) => {
-      amendmentIDs.set(sha512Half(Buffer.from(name, 'ascii')), {
-        name,
-        deprecated,
-      })
+async function findSingleAmendment(amendment_id: string): Promise<void> {
+  for (const url of Object.values(NETWORKS_HOSTS)) {
```
ckniffen, line 144: I think a function with a name starting with `find` should return the thing fount.

src/shared/database/amendments.ts
```diff
@@ -10,77 +14,163 @@ const log = logger({ name: 'amendments' })
...
+/**
+ * Fetch amendments in voting that are not available in feature-all RPC.
+ *
+ * @returns Void.
+ */
+async function fetchMissingAmendments(): Promise<void> {
+  const voting = new Set<string>()
+  await query('ballot')
+    .select('amendments')
+    .then((res) => {
```
ckniffen, line 161: this logic could exist outside of the `then` as an await is already used.

src/shared/database/amendments.ts
pdp2121: There are only these fields in `feature`, of which only name is useful for us.

```
enabled | Boolean | Whether this amendment is currently enabled in the latest ledger.

name | String | (May be omitted) The human-readable name for this amendment, if known.

supported | Boolean | Whether the server knows how to apply this amendment. If this field is set to false (the server does not know how to apply this amendment) and enabled is set to true (this amendment is enabled in the latest ledger), this amendment may cause your server to be amendment blocked.
```

src/shared/database/amendments.ts
pdp2121: They seem to use the term Obsolete or Retired more often. Maybe we can fix in future ticket since I will need to make changes to Explorer as well

src/shared/database/amendments.ts
```diff
@@ -10,77 +14,163 @@ const log = logger({ name: 'amendments' })
...
-    .digest('hex')
-    .toUpperCase()
-    .slice(0, 64)
+async function fetchSingleAmendment(
+  amendment_id: string,
+  url: string,
+): Promise<boolean> {
+  try {
+    log.info(`Updating amendment info for ${amendment_id}...`)
+    const client = new Client(url)
```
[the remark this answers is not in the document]
pdp2121: Each client connect to a different url so I'm not sure if there's a better way to do it here

src/shared/database/amendments.ts
pdp2121: I have made changes to resolve this

src/shared/database/amendments.ts
```diff
@@ -9,78 +13,142 @@ import { query } from './utils'
...
+    })
+
+    const featuresAll = featureAllResponse.result.features
+
+    for (const id of Object.keys(featuresAll)) {
+      amendmentIDs.set(id, {
+        name: featuresAll[id].name,
+        deprecated: RETIRED_AMENDMENTS.includes(featuresAll[id].name),
+      })
+      votingAmendments.delete(id)
```
achowdhry-ripple, line 85: Why are we removing amendments from the voting list?

achowdhry-ripple: nvm, understand

pdp2121: If an amendment info has been retrieved for amendments in voting I will just remove it so the next loop will not have to look up that info again

src/shared/database/amendments.ts
```diff
@@ -9,78 +13,142 @@ import { query } from './utils'
...
+    }
+
+    // Some amendments in voting are not available in feature all request.
+    // This loop tries to fetch them in feature one.
+    for (const amendment_id of votingAmendments) {
+      await fetchSingleAmendment(amendment_id, client)
```
achowdhry-ripple, line 91: nit: its a little confusing to call these fetch methods that internally set the cache of amendments as well. Would prefer if fetchsingleAmendment returned the one amendment and the setting happened in the parent call, but don't feel too strongly about it overall

src/shared/database/amendments.ts
pdp2121: Just renamed the variable to avoid confusion

src/shared/database/amendments.ts
pdp2121: Yes I agree. Let me fix it

achowdhry-ripple (approved):
LGTM -- for code readability would recommend differentiating between fetch methods and ones that actually set the cache as well, but logic looks solid otherwise

pdp2121:
> LGTM -- for code readability would recommend differentiating between fetch methods and ones that actually set the cache as well, but logic looks solid otherwise

Fixed that in the latest commit

anissa-ripple (approved):

PR #298, opened 2025-01-22
fix amendments backtracking

pdp2121:

There are 2 issues with amendment backtracking and status retrieval:
- `ledger` rpc for API v2 with `transactions` and `expand` enabled does not match `LedgerResponseExpanded` field on xrpl.js. Use v1 for now.
- s2 Clio often reached rate limit when backtracking amendments for mainnet. Use xrpl-cluster instead.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/connection-manager/wsHandling.ts
+++ b/src/connection-manager/wsHandling.ts
@@ -264,11 +264,15 @@ async function backtrackNetworkAmendmentStatus(
       index < currentLedger;
       index += 256
     ) {
+      // Use api V1 instead of the default (V2) since we don't have the correct
+      // LedgerResponseExpanded type for V2
+      // TODO: change to V2 once the issue is fixed.
       const ledger: LedgerResponse = await client.request({
         command: 'ledger',
         transactions: true,
         ledger_index: index,
         expand: true,
+        api_version: 1,
       })
 
       await handleWsMessageLedgerEnableAmendments(
--- a/src/shared/database/amendments.ts
+++ b/src/shared/database/amendments.ts
@@ -39,7 +39,7 @@ const AMENDMENT_VERSION_REGEX =
   /\| \[(?<amendmentName>[a-zA-Z0-9_]+)\][^\n]+\| (?<version>v[0-9]*\.[0-9]*\.[0-9]*|TBD) *\|/u
 
 export const NETWORKS_HOSTS = new Map([
-  ['main', 'ws://s2.ripple.com:51233'],
+  ['main', 'wss://xrplcluster.com'],
   ['test', 'wss://s.altnet.rippletest.net:51233'],
   ['dev', 'wss://s.devnet.rippletest.net:51233'],
 ])


src/shared/database/amendments.ts
```diff
@@ -39,7 +39,7 @@ const AMENDMENT_VERSION_REGEX =
   /\| \[(?<amendmentName>[a-zA-Z0-9_]+)\][^\n]+\| (?<version>v[0-9]*\.[0-9]*\.[0-9]*|TBD) *\|/u
 
 export const NETWORKS_HOSTS = new Map([
-  ['main', 'ws://s2.ripple.com:51233'],
+  ['main', 'wss://xrplcluster.com'],
```
mvadari, line 42: Why is this changed?

pdp2121: We often got rate limit while looping and sending `ledger` and `tx` RPC to clio (not sure if it's because of the recent issue) to backtrack amendments. We can either change this to xrpl-cluster completely or use it as a backup.

mvadari: Has this been discussed with the Clio folks?

pdp2121: `RippledError(You are placing too much load on the server., {\"error\":\"slowDown\",\"error_code\":10,\"error_message\":\"You are placing too much load on the server.\",\"status\":\"error` This happens once in a while with Clio

pdp2121: Just posted my question on their channel

src/connection-manager/wsHandling.ts
```diff
@@ -264,11 +264,15 @@ async function backtrackNetworkAmendmentStatus(
       index < currentLedger;
       index += 256
     ) {
+      // Use api V1 instead of the default (V2) since we don't have the correct
+      // LedgerResponseExpanded type for V2
+      // TODO: change to V2 once the issue is fixed.
```
mvadari, line 269: I recommend linking to the xrpl.js issue

ckeshava: Hello, do you remember the details of the issue pertaining to API v2 ? We are able to use the API v2 in the `Ledger` request for the latest validated ledger (a few lines above). I'm trying to understand why that would be a problem for other ledger indices.

The [response format](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/ledger-methods/ledger#response-format) of the `Ledger` command appears to be compatible with the VHS code [here](https://github.com/ripple/validator-history-service/blob/51d5db952bb7a776651e7bb67a5fe416195b6d6e/src/connection-manager/wsHandling.ts#L189)

src/shared/database/amendments.ts
ckeshava: I'm surprised that the testnet and devnet haven't thrown similar "slowdown" errors, given that the amendments' status are expected to be similar between the main, dev and test networks.

pdp2121: We are doing this because we are only using the entry node (so one node only) per each network to retrieve live amendments on Flag + 1 ledger. The backtracking would be just in case the entry node is down, we will use Clio. We only fetch Flag +1 ledgers in the last 30 minutes so I dont think it send too many requests to Clio

src/connection-manager/wsHandling.ts
pdp2121: Yes we dont need v1 now

PR #312, opened 2025-03-25
Fix: Updating IP address to null doesn't update

achowdhry-ripple:

If a user tries to update an existing crawls DB entry with a non-null IP to have a null IP, the IP address does not update. This is because we use merge in our saveNode, which has the behavior of excluding undefined values during conflict resolution. As a result, we need to manually override with the null fields we would like to retain (essentially sanitize the new entry with nulls instead of undefined)

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/shared/database/index.ts
+++ b/src/shared/database/index.ts
@@ -56,10 +56,24 @@ export async function saveNode(node: Node): Promise<void> {
     const ledgersSplit = node.complete_ledgers.split(',')
     node.complete_ledgers = ledgersSplit[ledgersSplit.length - 1]
   }
+
+  const sanitizedNode = {
+    complete_ledgers: null,
+    complete_shards: null,
+    start: null,
+    ip: null,
+    port: null,
+    type: null,
+    server_state: null,
+    io_latency_ms: null,
+    load_factor_server: null,
+    ...node,
+  }
+
   query('crawls')
-    .insert(node)
+    .insert(sanitizedNode)
     .onConflict('public_key')
-    .merge()
+    .merge(sanitizedNode)
     .catch((err: Error) => log.error(err.message))
 }
 
--- a/src/shared/database/index.ts
+++ b/src/shared/database/index.ts
@@ -58,15 +58,8 @@ export async function saveNode(node: Node): Promise<void> {
   }
 
   const sanitizedNode = {
-    complete_ledgers: null,
-    complete_shards: null,
-    start: null,
     ip: null,
     port: null,
-    type: null,
-    server_state: null,
-    io_latency_ms: null,
-    load_factor_server: null,
     ...node,
   }
 


pdp2121:
Can you fix the test case? Should only overwrite if either ip or port is null

mvadari:
nit: the PR title should explain what it's fixing, not the name of the issue (it doesn't really make grammatical sense right now)

achowdhry-ripple:
@mvadari sorry, had prematurely made this PR to see the pipeline output, should have marked it as draft. It is ready now

.vscode/settings.json
```diff
@@ -25,7 +25,7 @@
         "enable": true
     },
     "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": true
```
ckeshava, line 28: Did you mean to exclude this update from the changelog? I'd prefer the linter run automatically, what do you think?

src/crawler/crawl.ts
```diff
@@ -233,7 +233,7 @@ class Crawler {
       }
 
       this.publicKeysSeen.add(normalizedPublicKey)
-      promises.push(saveNode(dbNode))
+      promises.push(saveNode({ ip: undefined, port: undefined, ...dbNode }))
```
ckeshava, line 236: why are the `ip` and `port` values set to `undefined`? The `node` object already contains the correct `ip`, `port` fields, as returned from the `Crawl.crawlNode` method.

test/crawler/crawler.test.ts
```diff
@@ -65,6 +66,31 @@ describe('Runs test crawl', () => {
...
+    await crawl('1.1.1.1')
+
+    Object.keys(network3.peers).forEach((peer: string) => {
+      nock(`https://${peer}:51235`)
+        .get('/crawl')
+        .reply(
+          200,
+          (network3.peers as Record<string, Record<string, unknown>>)[peer],
+        )
+    })
```
ckeshava, line 79: This snippet can be added into the `mock()` function on line 10. I prefer this because there is a `beforeAll()` function which invokes `mock()`. This could ease future development of test cases.

test/crawler/crawler.test.ts
```diff
@@ -65,6 +66,31 @@ describe('Runs test crawl', () => {
     expect(results).toContainEqual(network1.result[2])
   })
 
+  test('successfully updates ip and port to null', async () => {
+    await crawl('1.1.1.1')
```
ckeshava, line 70:
```suggestion
```

Invoking the crawl method once is sufficient.

test/crawler/fixtures/null-ip-three-node-crawl.json
```diff
@@ -0,0 +1,94 @@
+{
+  "peers": {
+    "1.1.1.1": {
+      "overlay": {
+        "active": [
+          {
+            "ip": null,
```
ckeshava, line 7: In the broader context of this PR, what does this test represent? If a peer designates the ip, port information of an active_node as `null`, then we should be able to save it into the data base. Is that correct?

ckeshava (changes requested):
This PR appears to contradict this [line of code](https://github.com/ripple/validator-history-service/blob/910744a4e39946f3ef0dbe83fed13e99cca1af41/src/crawler/crawl.ts#L238)

If the ip (or) the port information is `null`, we do not want to crawl that endpoint (or) save it into the data-base. Is my understanding correct? Can you give me some examples for exceptions to this idea?

ckeshava:
Here is the documentation page for the appropriate [peer_crawler request](https://xrpl.org/docs/references/http-websocket-apis/peer-port-methods/peer-crawler#response-format). The `ip` and `port` fields might be omitted from the response, but I don't expect them to be `null`/`undefined`. Are you planning to handle the former case?
image

.vscode/settings.json
achowdhry-ripple: The line right below this maintains that the litner runs automatically. As of the latest VSCode versions, this boolean setting is converted to "explicit" so it was bound to be changed at some point. Seems like it is okay to keep as long as people are updating their vs code to the latest version

https://stackoverflow.com/questions/77637621/vscode-workspace-settings-change-on-its-own

src/crawler/crawl.ts
achowdhry-ripple: This is a workaround. Essentially, the database knex operation "merge" used in saveNode does not automatically set unincluded fields to null. When crawling, if an IP is null, it is currently excluded as a field from the JSON object, which then results in the new ip = null not being updated into the database. With this logic, if dbNode has an ip or port it will overwrite the undefined ones I have set since they occur later. However if it does not (as related to the bug this PR solves specifically), ip and port will still appear as undefined fields explicitly rather than just not being included

test/crawler/crawler.test.ts
achowdhry-ripple: I cannot do that because I am changing the simulated nodes in network1 specifically for this test, as opposed to creating an actual 3rd network that all tests would use. The code you have highlighted iterated over network3 peers, which all map to the same IP addresses as network 1 so I am updating the nock endpoints with the new node values, which includes the newly null ip node

test/crawler/fixtures/null-ip-three-node-crawl.json
achowdhry-ripple: yes, this designates the response we would get when crawling and a peer was changed to a null IP and port value.

achowdhry-ripple:
> Here is the documentation page for the appropriate [peer_crawler request](https://xrpl.org/docs/references/http-websocket-apis/peer-port-methods/peer-crawler#response-format). The `ip` and `port` fields might be omitted from the response, but I don't expect them to be `null`/`undefined`. Are you planning to handle the former case? image

This PR fixes a bug with the former case. If a node changes from a non-null IP to a null IP, we want that change reflected in the DB. Currently, it doesn't update, which is why this issue was raised from a client

achowdhry-ripple:
@ckeshava I see that without the context this change can seem obscure, so I have changed some variable names and added comments to the test for additional clarity

test/crawler/fixtures/null-ip-three-node-crawl.json
pdp2121: I would prefer ip and port here to not be set instead of null to align with behavior described in Peer Crawler doc

pdp2121 (approved):

pdp2121:
LGTM outside of the small change to test case mentioned

pdp2121:
Also, might be worth reach out to Brett to know which node has become private to make sure that would be fixed with this PR

src/crawler/crawl.ts
ckeshava: If a server operator wants to hide their ip/port information, should we retain that information in the crawls table? In this case, why shouldn't we delete the row pertaining to this node's `public_key` ?

Without the ip address and port information, this row in the `crawls` is not useful. All the other columns of the crawls table (like `ws_url`, `server_state`, etc) are not useful because the said server is unreachable.

src/crawler/crawl.ts
```diff
@@ -233,7 +233,7 @@ class Crawler {
       }
 
       this.publicKeysSeen.add(normalizedPublicKey)
```
ckeshava, line 235: Shouldn't this be gated by an `if` condition? How do you ensure that you are not over-writing the valid ip/port endpoints with the NULL values? (I'm referring to those server operators who still want to be discovered by the community)

ckeshava: Oh wait, since you have defined these fields `ip: undefined, port: undefined` in the beginning of the object, they can be over-written with valid values, if they exist. that makes sense

src/crawler/crawl.ts
pdp2121: I think it would just mean that the server becomes private. ip is not the only useful information for crawls table since we also show other information on the Explorer Nodes page. It's just not gonna be used for connections

ckeshava: image
Without the `ip:port` address, all of the other information is irrelevant, because the other columns change with time. If the node is not queried periodically, then all the info listed in the Explorer's Nodes page will be stale.

Schema of the `crawls` table: https://github.com/ripple/validator-history-service/blob/main/ARCHITECTURE.md#crawls


image

> I think it would just mean that the server becomes private.

I understand that node-operators choose to use `peer_private` in their rippled configurations.

These are the other places where the `crawls` table is used, however the `validators`, `networks` and `manifests` tables are more authoritative sources of truths.

1. [getNetworkFromPublicKey](https://github.com/ripple/validator-history-service/blob/51d5db952bb7a776651e7bb67a5fe416195b6d6e/src/api/routes/v1/get-network.ts#L176): The association of `public_key` with `networks` is periodically refreshed in the `validators` table.
2. [handleHealth](https://github.com/ripple/validator-history-service/blob/51d5db952bb7a776651e7bb67a5fe416195b6d6e/src/api/routes/v1/health.ts#L16): This is not a useful API endpoint. The connected column of the crawls table is [not populated correctly](https://github.com/ripple/validator-history-service/issues/323).
3. [findInDatabase](https://github.com/ripple/validator-history-service/blob/51d5db952bb7a776651e7bb67a5fe416195b6d6e/src/api/routes/v1/nodes.ts#L64) and [getNodes](https://github.com/ripple/validator-history-service/blob/51d5db952bb7a776651e7bb67a5fe416195b6d6e/src/api/routes/v1/nodes.ts#L109) : We need to use an outer-join of `validators` (or) `manifest` table with the `location` table here.

The rest of the uses of the `crawls` table are all utilities for db operations, `crawl` (or) `connection`-management operations.

ckeshava: If this is out of scope of the current PR, I am happy to implement in a different PR. What do you guys think?

pdp2121: I dont think this is true `Without the `ip:port` address, all of the other information is irrelevant, because the other columns change with time. If the node is not queried periodically, then all the info listed in the Explorer's Nodes page will be stale.` Private node still share information with its peer, so we can still get its information updated from its peer. By being private only means that 1. we cannot explore its peer, 2. we cannot get a websocket connection. The remaining content please raise an issue or create a PR, I will be happy to review

test/crawler/fixtures/null-ip-three-node-crawl.json
Patel-Raj: @achowdhry-ripple The unit tests fail in both of these cases:

1. comment out the test named `successfully crawls 3 node network`
2. Remove ip and port fields altogether from null-ip-three-node-crawl.json file for 1.1.1.1 node.

ckeshava: @Patel-Raj 
> comment out the test named successfully crawls 3 node network

I had a similar observation. This occurs due to the set up of `nock` endpoints. If the first test is removed, then the set up in `beforeAll` is not conducive for the current test. 

This is not a problem with this code change. It appears to be an artifact of the additional `nock` endpoint on the URL.

src/crawler/crawl.ts
ckeshava: Suppose `A` is behind a firewall and it has configured `B` as its `private_peer`. Hence, `A` is behind a firewall, whereas `B` is accessible by the world.

> Private node still share information with its peer

I'm confused because this contradicts the documentation. The [highlighted section](https://xrpl.org/docs/references/http-websocket-apis/peer-port-methods/peer-crawler#response-format) of the documentation states that if a node is configured as a `private_peer`, then its information is not included in the `peer_crawler` API response. So, `A` will not share information about its (say) forward proxies (`B`).

Are you suggesting the other direction of communication? If `B` shares information about `A`, doesn't it remove the "privacy" aspect of `A` ? Am I missing something here?

image

achowdhry-ripple:
Team, have fixed the issues. First I separated the nocks into each test in crawl.test.ts, rather than have all of them in beforeAll. this made the behavior easier to follow since they are declared as needed. Next, I added a param to the saveNode method since when crawling this_node, if there is not an IP passed in so we do not want to overwrite to null (we assume it was retrieved from a peer). In the peer iterations, we do want to set to null

test/crawler/fixtures/null-ip-three-node-crawl.json
achowdhry-ripple: this has now been fixed, and I separated the nocks per test for clarity.

src/shared/database/index.ts
```diff
@@ -49,17 +49,25 @@ export async function getNetworks(): Promise<Network[]> {
...
+export async function saveNode(
+  node: Node,
+  shouldOverwriteNull: boolean,
+): Promise<void> {
   if (node.complete_ledgers && node.complete_ledgers.length > 255) {
     const ledgersSplit = node.complete_ledgers.split(',')
     node.complete_ledgers = ledgersSplit[ledgersSplit.length - 1]
   }
+
+  const sanitizedMergeNode = shouldOverwriteNull
```
pdp2121, line 64: Maybe we can just set ip and port to null in the dbNode in `promises.push(saveNode(dbNode, true))` above? Then we don't need to modify saveNode at all?

achowdhry-ripple: I had it this way at first but it didn't work because then it always set the this_node IP to null, even though it might have already had an IP in a previous crawl from a different node.This resulted in a lot of nodes with null IP and port even though it wasn't actually the case. Thats why I added the variable to overwrite or not. Also IP and port do not allow for null by typing, so it would make the typing a bit more messy

src/crawler/crawl.ts
achowdhry-ripple: This conversation now seems out of scope for this PR. Since this PR specifically addresses a customer request, @ckeshava could you move this conversation to a new issue/PR?

ckeshava: @achowdhry-ripple  I still feel that a better solution (for this issue) is to remove the entry from the `crawls` table, instead of setting the `ip`/`port` information to `NULL`.

I don't mean to stall this PR, I'll watch out for any edge-cases that are not covered by my proposal.

pdp2121: The highlighted section only meant that the ip field will be omitted if the peer is private. Other info will still be available. So not sure why we should remove the row, the client that asked for this only asking for ip/port to be removed when their node became private, which likely means other info would still be available

src/crawler/crawl.ts
```diff
@@ -233,7 +234,9 @@ class Crawler {
...
-      promises.push(saveNode(dbNode))
+
+      // If a peer has the node IP as null, we want to overwrite in the database
+      promises.push(saveNode(dbNode, true))
```
Patel-Raj, line 239: @achowdhry-ripple Can you please check if only this change will fix the issue as suggested by @pdp2121 as well. 
We can remove boolean from the saveNode method and saveNode remains the same as its in main branch?

```suggestion
      promises.push(saveNode({ ip: null, port: null, ...dbNode}))
```

achowdhry-ripple: @Patel-Raj This would ruin the typing for the Node object, and require me to change `ip` and `port` to allow for null which breaks other parts of the code for me that reference these parameters. Personally I found this more messy and would rather update `saveNode` than make the typing here more opaque. Can discuss further if you feel strongly about this, or if I missed something and you were able to make this change a different way that doesn't require typing changes?

Patel-Raj: @achowdhry-ripple Ya, I had to change the type for ip and port and add a null check just below this line.

`promises.push(saveNode({ ip: null, port: null, ...dbNode }))`

```
ip?: string | null
port?: number | null
```

You are right, this also, gets a little messy but keeps the saveNode clean as its purpose is to save or update the node if it exists. No, strong opinions.

Patel-Raj (approved):

src/crawler/crawl.ts
achowdhry-ripple: true, I'd rather keep it like this for typing clarity in general, since this essentially adds a toggle for modifying the merge functionality in the db save which could be extensible in the future if there are more asks like this (other fields that should be nullable). If anyone strongly disagrees and would rather have it the other way, am open to discuss but will otherwise close out this thread and merge soon

src/crawler/crawl.ts
achowdhry-ripple: closing due to urgency and since this is nonbreaking. @ckeshava feel free to track this conversation in a new issue that can look into cleanliness of deleting a row vs updating to null

ckeshava (approved):

PR #325, opened 2025-04-02
Exponential backoff before WebSocket connection retry

Patel-Raj:

This change is being introduced to fix hourly agreement not being updated from xahau-main network. 

1. Added exponential backoff before WebSocket connection retries, so that the validator server is not swamped with connection requests and hopefully, our server's ip address is not blocked from any further requests.

2. Removed `connectionsInitialized` check from websockets `close` callback before retrying. Since, as per the current design of the code, if all the connections are not initialized and one of the connection (lets say wss://xahau.network/) closes , it won't be retried until next hour.

### Context of Change

To fix connection issue specifically for `wss://xahau.network/`

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

This issue occurs only in production. So, we will deploy this branch on staging and see if the intended fix do not cause any issue.

### DIFF
--- a/src/connection-manager/connections.ts
+++ b/src/connection-manager/connections.ts
@@ -1,5 +1,3 @@
-/* eslint-disable max-statements */
-/* eslint-disable max-depth */
 /* eslint-disable max-lines-per-function  -- Disable for this file with complex websocket rules. */
 import WebSocket from 'ws'
 import { LedgerEntryResponse } from 'xrpl'
@@ -33,8 +31,8 @@ const CM_INTERVAL = 60 * 60 * 1000
 const WS_TIMEOUT = 10000
 const REPORTING_INTERVAL = 15 * 60 * 1000
 const BACKTRACK_INTERVAL = 30 * 60 * 1000
-const BASE_DELAY = 1 * 1000 // 1 second
-const MAX_DELAY = 30 * 1000 // 30 seconds
+const BASE_RETRY_DELAY = 1 * 1000
+const MAX_RETRY_DELAY = 30 * 1000
 
 // The frequent closing codes seen so far after connections established include:
 //  1008: Policy error: client is too slow. (Most frequent)
@@ -54,6 +52,7 @@ let cmStarted = false
  * @param retryCount - Retry count for exponential backoff.
  * @returns A Promise that resolves to void once a connection has been created or timeout has occured.
  */
+// eslint-disable-next-line max-params -- Required here
 async function setHandlers(
   ws_url: string,
   ws: WebSocket,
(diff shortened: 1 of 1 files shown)


Vjkhannaripple (approved):
approved for deploying in staging

src/connection-manager/connections.ts
```diff
@@ -222,12 +255,12 @@ async function createConnections(): Promise<void> {
   })
 
   const promises: Array<Promise<void>> = []
-  connectionsInitialized = false
+
   nodes.forEach((node: WsNode) => {
     promises.push(findConnection(node))
   })
   await Promise.all(promises)
```
ckeshava, line 262: All these promises will resolve within a specified amount of time. The `findConnection` (this method is defined in this file) method uses `WS_TIMEOUT` on all the web-sockets. 

Regarding the connection attempts to xahau.network endpoint: how does the `connectionsInitialized` variable affect connection retries? The retry of connections to xahau.network is not gated anywhere in the code, except for the initial 10 seconds of the service.

src/connection-manager/connections.ts
```diff
@@ -107,41 +123,58 @@ async function setHandlers(
...
         )
-        if (CLOSING_CODES.includes(code)) {
-          log.info(
-            `Reconnecting to ${ws.url} on ${networks ?? 'unknown network'}...`,
-          )
+        // Clean up the old Websocket connection
+        connections.delete(ws.url)
+        ws.terminate()
+        resolve()
+
```
ckeshava, line 144:
```suggestion
void saveNodeWsUrl(ws.url, false)
```

Since the websocket is closed, that should be recorded in the database. If it reopens successfully, then there will be another update to the database inside `ws.on('open', ...)`

ckeshava: This is missing in the original code base too

src/connection-manager/connections.ts
Patel-Raj: If any validator refuses to open up a connection before `connectionsInitialized = true` this statement is executed, the current code will not retry to open another websocket connection due to this check `if (connectionsInitialized) {`. It might impact other websocket connections as well if they close before all the promises resolve. This is a hypothesis based on the current code, since its not being reproduced anywhere but production.

ckeshava (approved):
These changes add useful log-statements, thanks. However, it still falls short on one aspect: 

*We want to understand the health of a web-socket connection.* This will help us determine if the connection to xahau-main is receiving validations, manifests, etc regularly.

Currently, we don't have any way to evaluate the web-socket connections' reliability (except for digging through the logs). I suggest that we add another column in the `crawls` table titled `retries_in_last_hour` (or) `last_ping_time` (or) `closure_error_code`. Any of these values will tell us the last-seen status of the web-socket connection. It is straight-forward to populate these values too -- We can fill them inside `ws.on('close', ...)` (or) `ws.on('error', ...)`.

Unless xahau.network blocks the connection requests from VHS instance in the staging area, it is difficult to evaluate whether our modifications to the code are useful. If the servers do not resist any connections, then the `ws.on('close/error', ...)` codepath is never executed.

pdp2121 (approved):

src/connection-manager/connections.ts
Patel-Raj: This line causes deadlock since we swamp the database with many closed connection. The suggestion to add new column/table to track the health of the websocket connection is a good idea. We can store those health periodically.

PR #328, opened 2025-04-03
fix: remove db access per validation for fee data

pdp2121:

The validator table in database is being queried every time there's a new validation to retrieve fee data, which caused this error:
`KnexTimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?`
We can instead cache this validator-network data and update it every hour to reduce number of queries.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -123,10 +123,10 @@ function isPreceedingFlagLedger(ledger_index: string): boolean {
 }
 
 /**
- * Finds network name from chain id
+ * Finds network name from chain id.
  *
  * @param chain - A chain object.
- * @returns string.
+ * @returns String.
  */
 async function getNetworkNameFromChainId(chain: Chain): Promise<string> {
   let id = chain.id
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -26,7 +26,7 @@ import chains from './chains'
 
 const log = logger({ name: 'agreement' })
 
-const AGREEMENT_INTERVAL = 0.5 * 60 * 1000
+const AGREEMENT_INTERVAL = 60 * 60 * 1000
 const PURGE_INTERVAL = 10 * 60 * 1000
 
 /**
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -153,6 +153,10 @@ class Agreement {
     for (const chain of agreementChains) {
       const ledger_hashes = chain.ledgers
 
+      log.info(
+        `Agreement: ${chain.id}:${Array.from(chain.validators).join(',')}`,
+      )
+
       for (const signing_key of chain.validators) {
         promises.push(
           this.calculateValidatorAgreement(
--- a/src/connection-manager/agreement.ts
+++ b/src/connection-manager/agreement.ts
@@ -17,14 +17,16 @@ import {
   ValidationRaw,
   ValidatorKeys,
   Ballot,
+  Chain,
 } from '../shared/types'
+import { getLists, overlaps } from '../shared/utils'
 import logger from '../shared/utils/logger'
 
 import chains from './chains'
 
 const log = logger({ name: 'agreement' })
 
-const AGREEMENT_INTERVAL = 60 * 60 * 1000
+const AGREEMENT_INTERVAL = 0.5 * 60 * 1000
 const PURGE_INTERVAL = 10 * 60 * 1000
 
 /**
@@ -121,6 +123,30 @@ function isPreceedingFlagLedger(ledger_index: string): boolean {
 }
 
 /**
+ * Finds network name from chain id
+ *
+ * @param chain - A chain object.
+ * @returns string.
+ */
+async function getNetworkNameFromChainId(chain: Chain): Promise<string> {
+  let id = chain.id
+  const lists = await getLists().catch((err) => {
+    log.error('Error getting validator lists', err)
+    return undefined
+  })
+
+  if (lists != null) {
+    Object.entries(lists).forEach(([network, set]) => {
+      if (overlaps(chain.validators, set)) {
+        id = network
+      }
+    })
+  }
+
+  return id
+}
+
+/**
  *
  */
 class Agreement {
@@ -153,8 +179,12 @@ class Agreement {
     for (const chain of agreementChains) {
       const ledger_hashes = chain.ledgers
 
+      const networkName = await getNetworkNameFromChainId(chain)
+
       log.info(
-        `Agreement: ${chain.id}:${Array.from(chain.validators).join(',')}`,
+        `Agreement: ${chain.id}:${networkName}:${Array.from(
+          chain.validators,
+        ).join(',')}`,
       )
 
       for (const signing_key of chain.validators) {
--- a/src/connection-manager/connections.ts
+++ b/src/connection-manager/connections.ts
@@ -27,6 +27,7 @@ const ports = [443, 80, 6005, 6006, 51233, 51234]
 const protocols = ['wss://', 'ws://']
 const connections: Map<string, WebSocket> = new Map()
 const networkFee: Map<string, FeeVote> = new Map()
+const validationNetworkDb: Map<string, string> = new Map()
 const CM_INTERVAL = 60 * 60 * 1000
 const WS_TIMEOUT = 10000
 const REPORTING_INTERVAL = 15 * 60 * 1000
@@ -60,11 +61,11 @@ async function setHandlers(
   isInitialNode = false,
   retryCount = 0,
 ): Promise<void> {
-  log.info(
-    `Initiated Websocket connection for: ${ws_url} on ${
-      networks ?? 'unknown network'
-    }`,
-  )
+  // log.info(
+  //   `Initiated Websocket connection for: ${ws_url} on ${
+  //     networks ?? 'unknown network'
+  //   }`,
+  // )
 
   const ledger_hashes: string[] = []
   return new Promise(function setHandlersPromise(resolve, _reject) {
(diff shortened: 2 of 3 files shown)


src/connection-manager/agreement.ts
```diff
@@ -153,6 +153,10 @@ class Agreement {
     for (const chain of agreementChains) {
       const ledger_hashes = chain.ledgers
 
+      log.info(
```
shichengripple001, line 156: can we log the network name?

Patel-Raj (approved):

src/connection-manager/wsHandling.ts
```diff
@@ -115,22 +111,17 @@ export async function handleWsMessageSubscribeTypes(
   networks: string | undefined,
   network_fee: Map<string, FeeVote>,
   ws: WebSocket,
+  validationNetworkDb: Map<string, string>,
 ): Promise<void> {
   if (data.type === 'validationReceived') {
     const validationData = data as ValidationRaw
     if (ledger_hashes.includes(validationData.ledger_hash)) {
```
ckeshava, line 118: This `ledger_hashes` acts as a "cache" when VHS receives validations relevant to a network. We only need to perform the database query when we receive the "validation" before the "ledger". (Check the handling of ledgerStream below)

src/connection-manager/connections.ts
```diff
@@ -230,13 +232,23 @@ async function findConnection(node: WsNode): Promise<void> {
...
+
 /**
  * Creates connections to nodes found in the database.
  *
  * @returns A promise that resolves to void once all possible connections have been created.
  */
 async function createConnections(): Promise<void> {
   log.info('Finding Connections...')
+  validationNetworkDb.clear()
+  await getValidationNetworkDb()
```
ckeshava, line 251: The `createConnections` method is executed every `CM_INTERVAL` (1 hour). I'm concerned about the *staleness* of the data in the `validationNetworkDb` in-memory map.

New entries into the `validator` table are inserted in the [saveValidators](https://github.com/ripple/validator-history-service/blob/363e3dae6976489d2a76128e9f541f9e9cf8672f/src/shared/database/index.ts#L231) method. The `saveValidator` method is invoked upon receiving [every new validation](https://github.com/ripple/validator-history-service/blob/363e3dae6976489d2a76128e9f541f9e9cf8672f/src/connection-manager/wsHandling.ts#L139).

This is a hypothetical scenario: When we start the VHS connection service, we are not aware of many validators. However, we populate the `validationNetworkDb` in-memory map and it will not be updated for the next hour.

In the meantime, we can receive validations from new validators which are not present in our `validationNetworkDb` map. In the current code, we will mis-place these validations into the wrong network.

We can remedy this situation by falling-back to do a fresh db access, if we don't have any entry in `validationNetworkDb` in-memory map. Alternatively, we can run the `createConnections` method much more frequently.

ckeshava (changes requested):
I have my concerns about this design. This introduces complexity and tech-debt into the codebase due to the poor performance of the database. 

Do we have numbers on how many accesses are being performed on the `validators` table? 

If the db performance is the bottleneck, we will run into a similar problem with the `crawls` table too, because it is read and written into very frequently. It is not sustainable for us to build in-memory-cache for each of these "hot" tables.

I prefer to explore the alternatives to the current db.

src/connection-manager/wsHandling.ts
Patel-Raj: Can we rely on the networks filed passed to `export async function handleWsMessageSubscribeTypes(
` to compute the fees? It would completely eliminate need to store the validationNetworkDb in-memory map and database access. In which case that networks fields would be inconsistent?

src/connection-manager/connections.ts
```diff
@@ -166,12 +168,12 @@ async function setHandlers(
...
-        `Websocket connection error for ${ws.url} on ${
-          networks ?? 'unknown network'
-        } - ${err.message}`,
-      )
+    ws.on('error', () => {
+      // log.error(
+      //   `Websocket connection error for ${ws.url} on ${
+      //     networks ?? 'unknown network'
+      //   } - ${err.message}`,
+      // )
```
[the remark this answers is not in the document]
Patel-Raj: I agree, will add those back.

src/connection-manager/wsHandling.ts
ckeshava: @Patel-Raj 
> Can we rely on the networks field passed to export async function handleWsMessageSubscribeTypes(  to compute the fees? It would completely eliminate need to store the validationNetworkDb in-memory map and database access. In which case that networks fields would be inconsistent?

Yes, this is the current fall-back mechanism of the code. This is the `else` condition if the database does not have the network entry.

> We are picking up the ws_urls to open up websocket connection from crawls table. And this piece of code makes sure that at least some network is set. So, I wanted to know if we can remove the networks check in the first place and assume that it would always be defined.



Are you suggesting that we use the crawls table to fetch this network information? We can use it, however it only changes the database read access from `validators` to the `crawls` table. It doesn't reduce the load on the db per se.

ckeshava: What do you think of this alternative? If the database query returns a KnexError, we can catch that and set `ledger_fee = null;`. Fetching the network_fee is the main purpose of this data-base query.

However, it is not worth jeopardizing the operation of the VHS. Users can always query the network fees by directly invoking ledger APIs.

Patel-Raj: > Yes, this is the current fall-back mechanism of the code. This is the `else` condition if the database does not have the network entry.
> 
Why do we need to fallback? Can't we just use the networks passed in the handleWsMessageSubscribeTypes method as source of truth?

> Are you suggesting that we use the crawls table to fetch this network information? We can use it, however it only changes the database read access from `validators` to the `crawls` table. It doesn't reduce the load on the db per se.

No, I am not suggesting to use crawls table as fallback. I am suggesting to pick only those connection where networks is not null when this query is fired. 
```

const nodes = await query('crawls')
    .select(['ip', 'ws_url', 'networks'])
    .whereNotNull('ip')
    .andWhere('start', '>', tenMinutesAgo)
```

Patel-Raj: > What do you think of this alternative? If the database query returns a KnexError, we can catch that and set `ledger_fee = null;`. Fetching the network_fee is the main purpose of this data-base query.
> 
> However, it is not worth jeopardizing the operation of the VHS. Users can always query the network fees by directly invoking ledger APIs.

Its a good idea to have a try catch clause, but it would bring back the original problem of swamping the database with too many read queries.

ckeshava: what do you think of this PR? https://github.com/ripple/validator-history-service/pull/329

ckeshava:
```const nodes = await query('crawls')
    .select(['ip', 'ws_url', 'networks'])
    .whereNotNull('ip')
    .andWhere('start', '>', tenMinutesAgo)
```

^^ How is this query helpful in reducing the reads into the `validators` table? This query is used to open new web-socket connections.

However, we are overloading the database in the `ws.on('message', ...` callback function.

ckeshava: > Why do we need to fallback? Can't we just use the networks passed in the handleWsMessageSubscribeTypes method as source of truth?

It is possible for one rippled validator to be configured for multiple networks, for instance `main` and `dev`. Look at this [comment](https://github.com/XRPLF/rippled/blob/f608e653ca3cc1e410625ebaccf276f9f447e1a5/cfg/rippled-example.cfg#L833) for more info.

In such cases, I suspect one websocket connection could serve ledgers, validations pertaining to multiple networks. That is the only reason why we cannot use the `networks` input parameter to this function.

ckeshava: > But, https://github.com/ripple/validator-history-service/pull/329 will not go to fetch the network from database, it could remain undefined.

The networks value is not undefined. We are using the `validation` to read the `networks` value with a fall-back on the function input parameter.

I have identified one scenario where this code does not work correctly, however I don't know of many such cases. I believe its a worthy optimization to get the VHS back on track.

Patel-Raj: The validation gets its network from the function input parameter.
```

if (ledger_hashes.includes(validationData.ledger_hash)) {
    }
```

ckeshava: Here is the original interface:
```
interface ValidationRaw {
  flags: number
  full: boolean
  ledger_hash: string
  ledger_index: string
  master_key: string
  signature: string
  signing_time: number
  type: string
  validation_public_key: string
  server_version?: string

  networks?: string

  amendments?: string[]
  base_fee?: number
  reserve_base?: number
  reserve_inc?: number
  ledger_fee?: FeeVote

  // The validation_public_key is the same as the signing_key in StreamManifest
}
```

`validationData` does implicitly contain an optional network field. If this `ledger_hash` hasn't been encountered in recent hashes (as evidenced by the if-condition), then we are not using the `networks` function parameter.

src/connection-manager/connections.ts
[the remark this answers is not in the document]
pdp2121: I agree. We should just removed the log in websocket.close, since we are trying every possible ports to find one that can be connected, so printing it every single time will not be ideal

src/connection-manager/connections.ts
Patel-Raj: I added a fallback to database if we don't find in cache.

pdp2121: The `validationNetworkDb` map is only used to fetch fee voting data in case the current validation ledger_hash is not seen in the previous 10 hashes. It is not being used as source of truth for `networks`

src/connection-manager/connections.ts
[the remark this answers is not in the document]
ckeshava: The logs are helpful to understand if a web-socket endpoint has repeatedly denied connection attempts the VHS. Can we find out the connection closure reasons in any other way?

It does insert a log of log entries, but I can't think of a better way to measure the health of web-sockets.

src/connection-manager/wsHandling.ts
pdp2121: @Patel-Raj we cannot use the `networks` field passed in to populate the networks for validations since sometimes a node from one network can receive validation message on different networks. That was why we had to use past 10 ledger hashes as source of truth

ckeshava (approved):
The network_fee/ledger_fee column in the validators table might be stale for at most 1 hour. However, that will be rectified in the next update of the `validators` table.

PR #329, opened 2025-04-03
fetch network info from the validation

ckeshava:
The PR aims to reduce reading from the `validators` table. Since validation is the most recent source of truth, it must have the latest information about the `network`. If that field is not present, we can rely on the web-socket connection's metadata.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
Tests need to be added to verify the correctness. Furthermore, we need to measure the impact on the db performance with this change. This is planned for a future release.

### DIFF
--- a/src/connection-manager/wsHandling.ts
+++ b/src/connection-manager/wsHandling.ts
@@ -9,18 +9,13 @@ import {
 import { AMENDMENTS_ID } from 'xrpl/dist/npm/models/ledger'
 import { LedgerResponseExpanded } from 'xrpl/dist/npm/models/methods/ledger'
 
-import {
-  query,
-  saveAmendmentStatus,
-  saveAmendmentsStatus,
-} from '../shared/database'
+import { saveAmendmentStatus, saveAmendmentsStatus } from '../shared/database'
 import {
   NETWORKS_HOSTS,
   deleteAmendmentStatus,
 } from '../shared/database/amendments'
 import {
   AmendmentStatus,
-  DatabaseValidator,
   FeeVote,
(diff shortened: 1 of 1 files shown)


src/connection-manager/wsHandling.ts
```diff
@@ -123,14 +118,7 @@ export async function handleWsMessageSubscribeTypes(
...
     // Get network of the validation if ledger_hash is not in cache.
-    const validationNetworkDb: DatabaseValidator | undefined = await query(
-      'validators',
-    )
-      .select('*')
-      .where('signing_key', validationData.validation_public_key)
-      .first()
-    const validationNetwork =
-      validationNetworkDb?.networks ?? validationData.networks
+    const validationNetwork = validationData.networks ?? networks
```
Patel-Raj, line 121: As per https://github.com/ripple/validator-history-service/pull/328#discussion_r2027657024

I think we should fallback to the in-memory cache -> database to get the network as you suggested here https://github.com/ripple/validator-history-service/pull/328#discussion_r2027429145

ckeshava: Yes, that is also an acceptable solution. However, I'm wary of the larger untested code changes in that PR. I will also note one other (albeit rare) possibility:

The in-memory cache does not have any "invalidation scheme". The network corresponding to a validator's `public_key` could be changed after its admission into the in-memory cache. However, this change will not be registered until at least `CM_INTERVAL`.  This would be a very rare occurence, but it is something we need to consider for the future.

ckeshava:

Issue #67, opened 2022-12-19
Store `ledger_hash` with validators


ckniffen:
Right now we store the last ledger index we have validations for.  We should additionally store the `ledger_hash`.  This is important because it can be used to see if it is out of sync.

This will also clean up the validators page in the explorer which doesn't apply ledger hash color until the next ledger is validated.

ckeshava:
I personally prefer the `ledger_index` values to check if a node is out-of-sync. Isn't it hard to read the `ledger_hash` values?

ckeshava:
> Why not both? The Explorer needs both.

I understand, I'm trying to think of use cases where humans use `ledger_hash` values.

mvadari:
> > Why not both? The Explorer needs both.
> 
> I understand, I'm trying to think of use cases where humans use `ledger_hash` values.

Different networks may be on roughly the same `ledger_index` value, but they'll have different `ledger_hash` values. e.g. if Testnet and Devnet are restarted at around the same time.

ckeshava:
> Different networks may be on roughly the same ledger_index value, but they'll have different ledger_hash values. e.g. if Testnet and Devnet are restarted at around the same time.

That is true. However, on what occasions will developers see validated ledgers from different networks? Does this happen in practice?

The examples that I can think of -- WebSocket `ledger` streams, JSON-RPC requests of the `ledger` command etc are targeting one specific node. I'm guessing that users are aware of the `ValidatorList` used by such nodes (hence their "network" is well-known to the public).

mvadari:
It happens sometimes when a node is switching networks and still has peers from the old network.

PR #331, opened 2025-04-15
add `ledger_hash` values in `/network/validators` endpoint

ckeshava:

This PR aims to fix https://github.com/ripple/validator-history-service/issues/67. This is an old issue, I'm not sure if its still required, I did this for fun.

### Context of Change

The `validators` table already contains the `ledger_hash` column in its schema. This PR merely updates the external API endpoint at `/network/validators`.

This PR does not introduce any breaking changes. It adds a new field in one of the API responses.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

A new file that tests the `/network/validators` API endpoint is introduced in this PR.

## Future Tasks
I don't like the return format of  `/network/validators` API endpoint. It appears to cherry-pick most of the fields from the `validators` and `ballot` table, however some of the fields like `ledger_hash` appear to have fallen through the cracks.

[This file](https://github.com/ripple/validator-history-service/blob/51d5db952bb7a776651e7bb67a5fe416195b6d6e/src/api/routes/v1/validator.ts#L109) contains many examples of hand-maintained user-defined structs.

Is there a current/historical reason for these decision? The code is much more maintainable if return *all* the columns of a table in the API endpoint. The end-users are well equipped to filter for useful columns.


src/api/routes/v1/validator.ts
```diff
@@ -185,6 +188,7 @@ async function formatResponse(resp: dbResponse): Promise<ValidatorResponse> {
     server_version: resp.server_version,
     networks: resp.networks,
     current_index: Number(resp.current_index),
+    ledger_hash: resp.ledger_hash,
```
pdp2121, line 191: `current_hash` or `current_ledger_hash` would be more informative imo. We can just fix on API level here

ckeshava: I don't mind changing it, however users will not find any documentation on XRPL.org for `current_hash` (or) `current_ledger_hash`.

If you feel strongly, I can change it.

ckeshava: Furthermore, the [`validators` table also has a column](https://github.com/ripple/validator-history-service/blob/main/ARCHITECTURE.md#validators) with the name `ledger_hash`. It would be better to use the same name across all the tables, otherwise it is confusing.

Patel-Raj (approved):

src/api/routes/v1/validator.ts
pdp2121: I'm not suggesting we change name of table, but on the API result there's a `current_index` so I just want to call `ledger_hash` as `current_hash` on the API front so that it makes more sense when users see it. Will leave the call to you then since this is minor

pdp2121 (approved):

src/api/routes/v1/validator.ts
ckeshava: I see. I was not aware of the current_index API result, I agree that this will be an inconsistency. 👍 
However, I believe it is better to be consistent with the documentation on xrpl.org, instead of the VHS API. I'd prefer to retain this name, if that's alright.

Thanks for explaining the difference 👍

pdp2121:
@ckeshava you will need to rebase for CI to work

ckeshava:
thanks @pdp2121

PR #337, opened 2025-04-28
Expose WebSocket connection count as prometheus metrics and some refactors

Patel-Raj11:

This PR accomplishes the following two tasks:

1. Deprecate `ws_url` and `connected` columns from `crawls` table and add `connection_health` to capture their true values.
2. Expose number of active WebSocket connections for a network in prometheus exposition format.

### Context of Change

1. `ws_url` and `connected` columns from `crawls` table were not capturing all the active WebSocket connections since, we were creating additional WebSocket connections from using `entry` column of `networks` table as well. 
This change, adds a `connection_health` table that captures all the WebSocket connections and their status update time. It also, fixes the bug of `connected` column of `crawls` remaining stale even after loosing a WebSocket connection.
2. Exposes `/metrics/:network` that can be queried by Grafana to capture the number of connected nodes for each network and configure alerts.
3. Refactor the code to use xrpl.js library to fetch EnableAmendment Transaction and Amendments ledger entry as opposed to listening WebSocket events in `setHandlers` in order to make `setHandlers` more cleaner.

### Type of Change

- [x] Refactor (non-breaking change that only restructures code)
- [x] Documentation Updates
- [ ] Release

## Before / After

No change in the working on validator-history-service

## Test Plan

Added tests in /tests directory to check for /metrics/:network endpoint, database updates, selects, and utils methods that were introduced by this change.
I will deploy this branch on staging to check its working on stating before releasing on production.


mvadari:
Should these metrics only be exposed with a token of some sort, so they're not public?

Patel-Raj11:
> Should these metrics only be exposed with a token of some sort, so they're not public?

These metrics are similar to what we had [before](https://github.com/ripple/validator-history-service/blob/main/src/api/routes/v1/health.ts) but now, we are making them network specific and more accurate. I think they can remain public as they are health check metrics. Do you see a particular reason to make this endpoint authenticated?

pdp2121:
I would prefer these metrics to be tracked via Slack compared to exposing it publicly via API, since this is being used for internal monitoring purpose

pdp2121:
Would be something similar to what we do with VCR:
https://github.com/xpring-eng/validation-count-reporter

Patel-Raj11:
> I would prefer these metrics to be tracked via Slack compared to exposing it publicly via API, since this is being used for internal monitoring purpose

@pdp2121 

These metrics are going to be consumed by Grafana since they are exposed in prometheus exposition format for health checks and in-turn we will receive slack alerts from Grafana (as we do in #ripplex-alerting-dge-non-prod slack channel). In my opinion this is better approach since we will have decoupled alerting and application logic and we can move to a different alerting application like PagerDuty without modifying application logic at all.

However, if exposing the number of connected nodes as public API is a concern, then are we going to deprecate /health endpoint as well? Since, that endpoint has been exposing the same metric but in JSON format and for all nodes.

I can create separate PR if we opt for push based approach to send messages to slack since it would be cleaner to review it separately.

pdp2121:
`These metrics are going to be consumed by Grafana since they are exposed in prometheus exposition format for health checks and in-turn we will receive slack alerts from Grafana (as we do in #ripplex-alerting-dge-non-prod slack channel). In my opinion this is better approach since we will have decoupled alerting and application logic and we can move to a different alerting application like PagerDuty without modifying application logic at all. Just my thoughts, what do you think?`

If that's the case then I'm okay with exposing this data publicly since we have already listed number of connected node in `health` endpoint and we don't show this endpoint on the info list. @mvadari please let us know if you have any concerns

src/api/routes/v1/index.ts
```diff
@@ -18,6 +18,7 @@ import handleValidatorReport from './validator-report'
 const api = createRouter()
 
 api.use('/health', handleHealth)
+api.use('/metrics/:network', handleWebSocketHealthMetrics)
```
ckeshava, line 21: Can you add documentation for this API endpoint inside the `ARCHITECTURE.md` file and `src/api/routes/v1/info.ts` file?

test/connections/amendment.test.ts
```diff
@@ -77,4 +83,26 @@ describe('Amendments', () => {
     )
     expect(amendmentStatus[0].eta).toBe(null)
   })
+
```
ckeshava, line 86: Can you add unit tests for the other new methods pertaining to handling the amendments?

src/shared/database/setup.ts
```diff
@@ -55,6 +57,18 @@ async function setupCrawlsTable(): Promise<void> {
...
+    await db().schema.alterTable('crawls', (table) => {
+      table.dropColumn('ws_url')
+    })
+  }
+
+  if (await db().schema.hasColumn('crawls', 'connected')) {
+    await db().schema.alterTable('crawls', (table) => {
+      table.dropColumn('connected')
+    })
+  }
```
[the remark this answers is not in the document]
Patel-Raj11: We are removing `ws_url` and `connected` columns from crawls table. Those were populated when we create a connection to the node using `vhs-connections` service. There are no implications deleting this data. These columns in `connections_health` will be populated once the service starts.

src/shared/types/index.ts
```diff
@@ -227,6 +227,21 @@ interface AmendmentInfo {
   deprecated: boolean
 }
 
+interface ConnectionHealth {
+  ws_url: string
+  public_key?: string
```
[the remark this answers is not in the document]
Patel-Raj11: `const networksDb = await getNetworks()
  networksDb.forEach((network) => {
    nodes.push({
      ip: network.entry,
      ws_url: '',
      networks: network.id,
    })
  })`
  
 The rows that are picked by this query don't have public key. And this was the primary reason to introduce `connection_health` table so that we can keep track of connections that don't have public key.

At one point I thought why not remove this query and all the connections that we create from it all together, but then we are left with only one node for Xahau Mainnet and that is not ideal (For now its Xahau Mainnet but could occur with Testnet or other side chains in future). Also, `crawls` table is not a good place to store connection health, so having optional public_key helps us to track the entire state of connections.

src/connection-manager/connections.ts
```diff
@@ -100,129 +89,101 @@ async function setHandlers(
...
-        } with code ${code} and reason ${reason.toString('utf-8')}.`,
+        `Websocket closed for ${
+          ws.url
+        } on ${network} with code ${code} and reason ${reason.toString(
+          'utf-8',
+        )}.`,
       )
 
+      if (connections.get(ws.url)?.url === ws.url) {
+        connections.delete(ws.url)
```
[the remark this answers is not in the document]
Patel-Raj11: This is a good point. Let me try that out. Previously when I tried, it turned out that we were bombarding the database with to many requests on startup and getting deadlock. Since, most of the exhaustive tries were turning into either close/error callback before even and open callback and all those occur at once.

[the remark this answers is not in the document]
pdp2121: Yes I agree we dont need this anymore with new monitoring added. Good catch

src/connection-manager/wsHandling.ts
```diff
@@ -289,3 +320,70 @@ export async function backtrackAmendmentStatus(): Promise<void> {
...
+ *
+ * @param network - Network name.
+ * @param hostName - Hostname to connect.
+ * @returns Void.
+ */
+async function getAmendmentsFromLedgerEntry(
+  network: string,
+  hostName: string,
+): Promise<void> {
+  const allUrls: string[] = []
```
pdp2121, line 335: Can we add log at the beginning and ending of this process?

pdp2121: Same for other processes as well. Would be something like starting ... and finished ...

src/shared/database/index.ts
```diff
@@ -72,48 +73,52 @@ export async function saveNode(
...
 
+/**
+ * Get the list of nodes to establish WebSocket connection.
+ *
+ * @param sinceStartDate -- Date instance from which to retrieve data.
+ * @returns The list of nodes.
+ */
+export async function getNodes(sinceStartDate: Date): Promise<WsNode[]> {
+  return query('crawls as c')
+    .leftJoin('connection_health as ch', 'c.public_key', 'ch.public_key')
```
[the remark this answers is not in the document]
Patel-Raj11: I think due to reason mentioned in https://github.com/ripple/validator-history-service/pull/337#discussion_r2066702366 we can have only `ws_url` as primary key since, at one point in time there will be one connection for a `ws_url` since it contains IP and port information.

test/connections/amendment.test.ts
Patel-Raj11: I have added this test `save Amendments ledger_entry` which tests for `fetchAmendmentsFromLedgerEntry` to check the data returned from ledger_entry is of correct shape and gets saved properly. 

For `processEnableAmendmentTransaction` the data returned by ledger command is same as the existing tests that we have (e.g. ledgerResponseNoFlag, ledgerResponseGotMajority etc). So, it already gets tested as we are in turn calling the same method `handleWsMessageLedgerEnableAmendments`

There is no logic here in these methods. What else do you think can be tested here?

src/shared/database/setup.ts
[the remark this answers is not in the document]
pdp2121: Should not take more than couple minutes after redeployment. We already have the crawls table so whenever websocket connections established we should have data populated into this table.

I don't think we need to migrate data here. Just need to remove that column in `crawls` table in redeployment by calling a knex drop column query. The `connected` column reflect the current websocket connection, so no data is being lost here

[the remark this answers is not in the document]
pdp2121: The code in the setup file will run once whenever the service restarts just FYI

src/connection-manager/wsHandling.ts
```diff
@@ -143,27 +121,80 @@ export async function handleWsMessageSubscribeTypes(
...
+): void {
+  if (!network) {
+    return
+  }
+
+  if (!isFlagLedgerPlusOne(ledgerIndex)) {
+    return
+  }
+
+  // Already seen by some other node for this network. No need to save again.
```
pdp2121, line 155: Nice! I like this better than the initial node approach

src/shared/database/setup.ts
```diff
@@ -251,3 +265,16 @@ async function setupBallotTable(): Promise<void> {
     })
   }
 }
+
+async function setupConnectionHealthTable(): Promise<void> {
+  const hasConnectionHealth = await db().schema.hasTable('connection_health')
+  if (!hasConnectionHealth) {
+    await db().schema.createTable('connection_health', (table) => {
+      table.string('ws_url').primary()
+      table.string('public_key').references('crawls.public_key')
```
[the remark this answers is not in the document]
pdp2121: Should we make both ws_url and pubkey primary? Since if 2 nodes share the same ws_url, we still open 2 websocket connections, which would create 2 rows in this table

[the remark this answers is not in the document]
Patel-Raj11: @pdp2121 `ws_url` comprises of ip and port information. So two nodes will be serving from two different ports even if they run on a single machine. So `ws_url` would still be unique.

Moreover, composite primary key should not have a null column and public_key is a candidate null column for the entries fetched from networks table.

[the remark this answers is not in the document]
pdp2121: Makes sense to me

src/connection-manager/connections.ts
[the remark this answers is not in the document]
Patel-Raj11: Removing in-memory map and directly trying update the database if the connection closes, brings back this error:

`Failed to update connection status for ws://72.44.58.240:6006/: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?`

I am trying to increase [acquireconnectiontimeout](https://knexjs.org/guide/#acquireconnectiontimeout) to 2 minutes and see if it helps to resolve this.

.env.example
```diff
@@ -17,6 +17,7 @@ DB_HOST=
 DB_USER=
 DB_DATABASE=
 DB_PASSWORD=
+ACQUIRE_CONNECTION_TIMEOUT=
```
pdp2121, line 20: What's the use of this variable?

pdp2121, line 20: Why do we need this to be env variable instead of a constant?

src/api/routes/v1/info.ts
```diff
@@ -82,6 +82,17 @@ const info = {
       example:
         'https://data.xrpl.org/v1/network/amendments/vote/{network}/{identifier}',
     },
+    {
```
pdp2121, line 85: Since this is for internal use I don't think we need to expose it in info

src/connection-manager/connections.ts
```diff
@@ -100,129 +90,87 @@ async function setHandlers(
...
+        validationNetworkDb,
+        enableAmendmentLedgerIndexMap,
+      )
     })
     ws.on('close', async (code, reason) => {
       log.error(
-        `Websocket closed for ${ws.url} on ${
-          networks ?? 'unknown network'
-        } with code ${code} and reason ${reason.toString('utf-8')}.`,
+        `Websocket closed for ${
```
pdp2121, line 105: I'm not sure if we need this log now with your change. It seems to flood the logs atmm

src/connection-manager/connections.ts
```diff
@@ -100,129 +90,87 @@ async function setHandlers(
...
-      }
-      ws.terminate()
       resolve()
     })
     ws.on('error', (err) => {
       log.error(
-        `Websocket connection error for ${ws.url} on ${
-          networks ?? 'unknown network'
-        } - ${err.message}`,
+        `Websocket connection error for ${ws.url} on ${network} - ${err.message}`,
```
pdp2121, line 132: Same here. Since we try multiple ports for an url, this will raise lots of errors that has no important implication

src/connection-manager/connections.ts
Patel-Raj11: @pdp2121 This is only flooding the logs initially when the service starts up. At any other point in time it will give valuable information about when the connection was terminated and with what reason. We can keep these two logs for now and remove later once we don't see any value. What do you think?

src/shared/database/setup.ts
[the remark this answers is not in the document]
ckeshava: @Patel-Raj  I'm still confused
> I don't see a reason to have multiple public keys for a ws_url in connection_health

Can you explain the results of this query? 
```
SELECT ws_url, COUNT(public_key) AS shared_public_key_count FROM crawls WHERE ws_url IS NOT null GROUP BY ws_url HAVING COUNT(public_key) > 1;
```

image

This result indicates that each of these 176 `ws_url` entries are shared by multiple unique `public_key`s. Your proposed design will over-write the existing public_key information. Aren't we losing useful data in this process?

> Now, since we pick only the rows having recent start date (i.e less then 10 mins old) the other old rows for same ws_url are no more relevant and stale.

Am I understanding correctly that rows older than 10 minutes are not useful in the `crawls` table? I'm working on a different task for optimizing the space usage of the database. I think we should purge old data and keep the `crawls` table lean.

src/connection-manager/connections.ts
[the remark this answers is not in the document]
ckeshava: thanks @Patel-Raj. Sorry, I didn't mean to dump this task onto you. I'm happy to take a look at this issue in a later PR. This was a tangential suggestion, not directly related to your goal.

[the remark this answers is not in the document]
Patel-Raj11: @ckeshava This change suggested and [acquireconnectiontimeout](https://knexjs.org/guide/#acquireconnectiontimeout) increase to 2 minutes is already in the PR. Just in case you missed it to review.

test/connections/amendment.test.ts
Patel-Raj11: `checkAndHandleEnableAmendmentLedger` this method is just a refactor of the previous if conditions to make it cleaner and its not exported so I cannot test it in isolation. Also, the entire `setHandlers` is not tested  and I don't have clear idea on how to test it. We can take this task of writing tests for `setHandlers` this separate effort.

src/shared/database/setup.ts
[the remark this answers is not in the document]
Patel-Raj11: `176 ws_url entries` have port information along with their protocol and ip as seen in screenshot. So, one ws_url at a time can have only one public_key (thus node) actively listening to web socket connection.
 
Also, multiple public keys which we see are from different start times. You can pick one of the `ws_url` and query the crawls table to see the start time. So, they are not being used at the same time.

I don't think we loose any information.

Yes, rows older than 10 minutes are stale and of now use. We can write a cleanup job to remove those.

src/api/routes/v1/info.ts
pdp2121: I meant we don't show it in the info when users hit https://data.xrpl.org, but still keep it within architecture document

pdp2121 (approved):
Looks good to me other than the websocket closed discussion

Patel-Raj11:
> @Patel-Raj11 are you still having the branch ran on staging? Just want to check if it captures DeepFreeze correctly

Yes, past deploys had this [branch](https://github.com/xpring-eng/vhs-deploy/actions)

pdp2121:
> > @Patel-Raj11 are you still having the branch ran on staging? Just want to check if it captures DeepFreeze correctly
> 
> Yes, past [deploys](https://github.com/xpring-eng/vhs-deploy/actions) had this branch

Yes staging looks good to me

src/connection-manager/connections.ts
pdp2121: Sounds good to me

src/api/routes/v1/health.ts
```diff
@@ -13,11 +13,40 @@ export default async function handleHealth(
   res: Response,
 ): Promise<void> {
   try {
-    const count = (await query('crawls')
-      .countDistinct('ip')
+    const count = (await query('connection_health')
```
achowdhry-ripple, line 16: would this now exclude the private nodes?

Patel-Raj11: `ws_url` addresses in `connection_health` table is formed using ip address in `crawls` table. Yes, if a node is private it is excluded from being connected and thus is not reported in this count.

ARCHITECTURE.md
```diff
@@ -12,6 +12,7 @@ There are 3 folders in `src`, corresponding to the 3 processes that the VHS runs
 * `/`: Information about the endpoints.
 * `v1`
   * `/health`: A health check for the VHS. Returns the number of nodes that it is connected to.
+  * `/metrics`: A health check for the VHS. Returns the number of connected nodes for each network in prometheus exposition format.
```
achowdhry-ripple, line 15: how come we are keeping both endpoints? It seems like the new metrics endpoint is a more detailed version of the same info the health endpoint has

achowdhry-ripple (approved):
lgtm, I have limited context on this but the code makes sense to me overall -- nice job with this

test/connections/connection-health.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  test('update existing connection_health row', async () => {
+    const initialData: ConnectionHealth = {
+      ...data.initial,
+      status_update_time: new Date(data.initial.status_update_time),
+    }
+
+    const updatedData: ConnectionHealth = {
+      ...data.updated,
+      status_update_time: new Date(data.initial.status_update_time),
```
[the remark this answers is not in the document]
Patel-Raj11: Good catch, it didn't matter for the test case but I changed it to avoid any confusions.

ARCHITECTURE.md
Patel-Raj11: The new metrics endpoint returns data in prometheus exposition format and the /health returns data in JSON format. Since, it was public endpoint users of VHS might be consuming it. No harm in keeping both.

PR #367, opened 2025-08-07
fix: use xrpscan for amendment version

pdp2121:

xrpl.org is no longer providing amendment versions on their Known Amendment page.
Use XRPScan API to fetch this data instead

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,3 +27,6 @@ MAINNET_UNL=vl.ripple.com
 
 # Rippled Server with access to manifest command
 RIPPLED_RPC_ADMIN=
+
+# XRPScan public API, available at https://docs.xrpscan.com/api-documentation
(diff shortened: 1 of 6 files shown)


mvadari (approved):
There should be an issue for getting this info straight from Github (I feel like it should be possible to determine when an amendment was set to `Supporting::Yes` from the `features.macro` file and its git history)

pdp2121:
> There should be an issue for getting this info straight from Github (I feel like it should be possible to determine when an amendment was set to `Supporting::Yes` from the `features.macro` file and its git history)

Make sense. Not sure how Anurag fetched his data (manually or parsing the release notes from rippled), but yes I will create a ticket for it

PR #370, opened 2025-08-21
Store validated ledgers inside a persistent DB

ckeshava:

This PR aims to comprehensively track all the validated and missing ledgers, as recorded by the VHS. It also exposes suitable API endpoints to query this piece of information.

If you prefer the API endpoints (or) the DB schema differently for your use-cases, let me know.

The `validated_ledgers` table is pruned daily. I have not set up a pruning schedule for `missing_ledgers` table yet for debugging purposes. Presently, both these tables only track the mainnet for reducing the storage burden. If deemed necessary, this can be extended to other networks as well.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Here is an example of how I envision the usage of this PR changes:
**Debugging:** Inspect the `missing-ledgers` table. Note the `ledger_index`, `previous_ledger` and `previous_received_at` timestamps.
image

Use the information from the contextual rows in the `validated_ledgers` table to gather more information about the timestamp of the missing ledger. Use this timestamp to correlate events with the logs of `vhs-connections` pod. (Did the pod restart? Was there a prior unhandled error? etc)

**Source of Truth:** The `missing_ledgers` and `validated_ledgers` provide a comprehensive store of the VHS's data. It can be used to pinpoint erroneous behavior.

If one does not have access to the DB, this can be accomplished with the exposed API endpoints as well.

## Test Plan

Unit tests have been added for the DB tasks pertaining to `validated_ledgers` table. If we are planning to retain the `missing_ledgers` table in the production VHS instance, suitable tests can be added in the future.

## Future Tasks

In a future PR, I'd like to track all the incoming validations received by the VHS.


src/api/routes/v1/index.ts
```diff
@@ -14,11 +14,16 @@ import handleNetworks from './networks'
...
+import handleValidatedLedgers from './validated-ledgers'
+import handleMissingLedgers from './missing-ledgers'
 
 const api = createRouter()
 
 api.use('/health', handleHealth)
 api.use('/metrics', handleMonitoringMetrics)
+
+api.use('/validated-ledgers/:network', handleValidatedLedgers)
+api.use('/missing-ledgers/:network', handleMissingLedgers)
```
kuan121, line 26: Who will use these endpoints and how?

src/shared/database/validatedLedgers.ts
```diff
@@ -0,0 +1,82 @@
...
+import { query } from './utils'
+
+/**
+ * Inserts a validated ledger into the database, ignoring if it already exists.
+ *
+ * @param network - The network identifier.
+ * @param ledger - The ledger data to insert.
+ * @returns A promise that resolves when the insertion is complete.
+ */
+export async function insertValidatedLedger(
```
Patel-Raj11, line 12: I think it would be good to insert both ledger_time which is as per RIPPLE_EPOCH and time since Unix epoch so that can do count() queries on both and compare the data directly with Capsian.

Patel-Raj11: Apart from this, looks good for deployment on stg/dev env.

src/shared/database/validatedLedgers.ts
```diff
@@ -0,0 +1,82 @@
...
+  network: string,
+  ledger: StreamLedger,
+): Promise<void> {
+  await query('validated_ledgers')
+    .insert({
+      network,
+      ledger_hash: ledger.ledger_hash,
+      ledger_index: ledger.ledger_index,
+      ledger_time: ledger.ledger_time,
```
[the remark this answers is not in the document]
Patel-Raj11: We might receive that ledger late, so `received_at` does not depict `ledger_time`. I am not super concerned, since we will apply some cool date addition/subtraction logic when fetching the data if we want time with reference to Unix Epoch 😄.

I thought having it for easier queries, but not a big deal.

package.json
```diff
@@ -10,7 +10,7 @@
     "startCrawlerDev": "ts-node-dev --transpile-only ./src/crawler/index.ts | ./node_modules/.bin/bunyan ",
     "startApiDev": "ts-node-dev --transpile-only ./src/api/index.ts | ./node_modules/.bin/bunyan",
     "startConnectionsDev": "ts-node-dev --transpile-only ./src/connection-manager/index.ts | ./node_modules/.bin/bunyan",
-    "test": "jest",
+    "test": "jest --detectOpenHandles",
```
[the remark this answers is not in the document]
ckeshava: the additional flag helps me detect memory-leaks/unclosed `setInterval` invocations in the code.

src/api/routes/v1/index.ts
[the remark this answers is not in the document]
ckeshava: 2. I meant to use these API endpoints temporarily for the purposes of trouble-shooting. I don't know of any consumers who would like it in production. However, the `missing-ledgers` and `validated-ledgers` tables are themselves useful, even in production VHS systems.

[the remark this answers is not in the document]
ckeshava: 1. The XRPL blockchain processes 1 validated ledger every 3-4 seconds. Hence, XRPL produced a maximum of 24 * 3600 / 3 = 28,800 ledgers per day. Hopefully, this estimate will help the API users in estimating the `limit` value.

I don't mind removing these two API endpoints -- I only included them for ease of use.

src/shared/database/validatedLedgers.ts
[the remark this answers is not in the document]
Patel-Raj11: > Why should we care about when VHS receives a ledgerClosed message?
There were some considerations that chain_id logic might not work if the ledger/validation message is received late by VHS. I think let's have this, in case it might help in those analysis.

src/connection-manager/connections.ts
```diff
@@ -233,5 +234,49 @@ export default async function startConnections(): Promise<void> {
...
+async function pruneValidatedLedgersTable(): Promise<void> {
+  await query('validated_ledgers')
+    .where('received_at', '<', new Date(Date.now() - 1000 * 60 * 60 * 24 * 30))
+    .del()
+}
+
+// check for missing ledgers every minute
+const SCAN_PERIOD = 1000 * 60 * 5
+
+async function checkForMissingLedgers(): Promise<void> {
```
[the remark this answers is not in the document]
ckeshava: I've added unit tests here: 7c5c96f

src/shared/database/setup.ts
```diff
@@ -278,3 +280,34 @@ async function setupConnectionHealthTable(): Promise<void> {
...
+  if (!hasTable) {
+    await db().schema.createTable('validated_ledgers', (table) => {
+      table.string('network').notNullable()
+      table.string('ledger_hash').notNullable()
+      table.bigInteger('ledger_index').notNullable()
+      table.bigInteger('ledger_time').notNullable()
+      table.bigInteger('fee_base').notNullable()
+      table.bigInteger('reserve_base').notNullable()
+      table.bigInteger('reserve_inc').notNullable()
+      table.integer('txn_id')
```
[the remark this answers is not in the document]
ckeshava: yes, this has been a mistake in the VHS since many years. I wanted to keep this CL as small as possible. I can tackle this mistake in a different PR.

Let me know if you prefer that I fix this error in this PR

src/connection-manager/connections.ts
```diff
@@ -233,5 +234,49 @@ export default async function startConnections(): Promise<void> {
...
+    if (
+      parseInt(recentLedgers[i + 1].ledger_index, 10) !==
+      parseInt(currentLedger.ledger_index, 10) + 1
+    ) {
+      await insertMissingLedger({
+        network: 'main',
+        ledger_index: parseInt(currentLedger.ledger_index, 10) + 1,
+        previous_ledger_index: parseInt(currentLedger.ledger_index, 10),
+        previous_ledger_received_at: currentLedger.received_at,
+      } as MissingLedger)
```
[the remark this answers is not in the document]
ckeshava: aren't we over-optimizing the db operations here? it should be the responsibility of DB to batch/optimize its operations right? I don't mind batching it, but I'm wary of over-complicating the codebase.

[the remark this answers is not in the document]
kuan121: You raise a good point about over-optimization. My main concern was avoiding unnecessary network calls, since the VHS connection and database run on separate k8s pods and each write incurs a network call. Bulk or batch writes are an industry standard, but given the relatively low write volume in your case, it may not be strictly necessary. Feel free to skip it for now. That said, I don’t think the implementation would add much complexity—it could be as simple as:
```
const missingLedgersArray = []; // Collect all missing ledgers
db().bulkInsert(missingLedgersArray); // I can’t access the Knex docs right now, but I believe it supports bulk inserts
```

src/shared/database/validatedLedgers.ts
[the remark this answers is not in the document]
ckeshava: I have transformed the `ledger_time` column into UTC time-format in this commit: 986698f975c6191dccb3a1021e1bf22b8d10918e

src/api/routes/v1/index.ts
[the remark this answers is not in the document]
ckeshava: this API endpoint has been removed in recent version of the PR.

src/api/routes/v1/validated-ledgers.ts
```diff
@@ -0,0 +1,53 @@
...
+  req: Request,
+  res: Response,
+): Promise<Response> {
+  const network = req.params.network
+  const limitStr = req.query.limit as string | undefined
+  let limit: number | undefined
+
+  if (limitStr) {
+    limit = parseInt(limitStr, 10)
+    if (limit <= 0) {
```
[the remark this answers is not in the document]
ckeshava: this API endpoint has been removed in recent version of the PR.

ARCHITECTURE.md
```diff
@@ -198,5 +200,29 @@ This table keeps track of the WebSocket connection status for all networks.
...
+### `missing_ledgers`
+
+This table keeps track of missing validated ledgers detected across different networks.
+
+| Key                          | Definition                                                              |
+|------------------------------|-------------------------------------------------------------------------|
+| `network`                    | The network on which the ledger is missing.                             |
+| `ledger_index`               | The index of the missing ledger.                                        |
+| `previous_ledger_index`      | The index of the previous ledger that was received.                     |
+| `previous_ledger_received_at`| The time when the previous ledger was received by the VHS.              |
```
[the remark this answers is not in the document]
ckeshava: Hello Kuan, the `missing_ledgers` table is no longer a part of this PR. Let me know if you have any concerns about the `validated_ledgers` table.

ARCHITECTURE.md
```diff
@@ -198,5 +198,18 @@ This table keeps track of the WebSocket connection status for all networks.
 | `connected`          |Boolean denoting websocket connection status.                      |
 | `status_update_time` |Time when the connected column was updated.                        |
 
+### `validated_ledgers`
```
Patel-Raj11, line 201: Lets update this after all the comments and addressed.

src/connection-manager/connections.ts
```diff
@@ -233,5 +233,14 @@ export default async function startConnections(): Promise<void> {
     setInterval(() => {
       void backtrackAmendmentStatus()
     }, BACKTRACK_INTERVAL)
+
+    setInterval(pruneValidatedLedgersTable, 1000 * 60 * 60 * 24)
```
Patel-Raj11, line 237: 1. 1000 * 60 * 60 * 24 => Define a constant for this.
2. Lets run this job every day instead of every hour.

src/shared/database/validatedLedgers.ts
```diff
@@ -0,0 +1,90 @@
...
+export async function insertValidatedLedger(
+  network: string,
+  ledger: StreamLedger,
+): Promise<void> {
+  await query('validated_ledgers')
+    .insert({
+      network,
+      ledger_hash: ledger.ledger_hash,
+      ledger_index: ledger.ledger_index,
+      ledger_time: new Date(rippleTimeToUnixTime(ledger.ledger_time) * 1000),
```
Patel-Raj11, line 26: 1. There shouldn't be a need to multiply by 1000. See [this](https://github.com/ripple/validator-history-service/blob/b81d0608c03143078c8eaabecfc33fafc1517548/src/shared/database/update-amendments-from-json.ts#L73).
2. try catch is required here. With logging error message in catch.

src/shared/database/validatedLedgers.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * This method inserts validations associated with a ledger into the validated_ledgers table.
+ *
+ * @param ledger_hash - The hash of the ledger.
+ * @param ledger_index - The index of the ledger.
+ * @param validation_public_keys - The signing keys associated with the validations of this ledger.
+ * @param networkName - The name of the network.
+ * @returns A promise that resolves when the insertion is complete.
+ */
+export async function insertValidations(
```
Patel-Raj11, line 57: 1. If there is exactly one entry we update it. If there are 0 entries we log a message that you are already logging and insert the row as well. Last `else` clause should never occur but good to have it for logging purposes.
2. try catch is required here, otherwise DB operation failure will fail the actual source code flow.

src/shared/database/setup.ts
```diff
@@ -278,3 +279,23 @@ async function setupConnectionHealthTable(): Promise<void> {
...
+      table.bigInteger('ledger_index').notNullable()
+      // Note: The VHS transforms the value from Ripple-Epoch-Time to UTC format
+      table.dateTime('ledger_time').notNullable()
+      table.bigInteger('fee_base').notNullable()
+      table.bigInteger('reserve_base').notNullable()
+      table.bigInteger('reserve_inc').notNullable()
+      table.integer('txn_id')
+      table.specificType('validation_public_keys', 'TEXT[]')
+      table.dateTime('received_at').defaultTo(db().fn.now())
+      table.primary(['ledger_index', 'network', 'ledger_hash'])
```
Patel-Raj11, line 298: 1. I think composite primary key on `ledger_hash` and `network` would suffice.
2. Except columns involved in primary key, all other columns should be nullable. Since, we can receive an incorrect validation for which there is no ledger from ledger stream (the case the @pdp2121 was suggesting to capture), we want to insert that entry as well. More details on this below.

src/shared/database/validatedLedgers.ts
ckeshava: > If there are 0 entries we log a message that you are already logging and insert the row as well (to account for validation that we receive for which we never had a ledger message).

Regarding this case, if the VHS has not recieved the `ledgerClosed` message for a specific ledger, then I cannot insert the row. I do not have the necessary fields of the `StreamLedger` interface to be able to fill out these values at this stage. Executing a `log.error()` appears to be the best course of action here.

src/shared/types/index.ts
```diff
@@ -19,14 +19,15 @@ interface Ledger {
 
 interface StreamLedger {
```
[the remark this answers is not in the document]
ckeshava: No, the VHS source code was out-of-sync with the updates to the `subscribe` command response. I took this opportunity to revamp the source code.

src/shared/database/setup.ts
ckeshava: > I think composite primary key on ledger_hash and network would suffice.

There is a non-trivial possibility to reconstruct similar/identical `ledger_hash` values in a custom XRPL network (either due to byzantine actors (or) pure chance). I have not thought too much about this possibility. 
Since there is no disadvantage to have one more column in the composite primary_key of the DB, this is better for the safety/robustness of the VHS.

src/shared/database/validatedLedgers.ts
ckeshava: thanks, fixed in 84740c0

src/connection-manager/connections.ts
ckeshava: 84740c0

Patel-Raj11: nit: I think this got missed.

src/connection-manager/wsHandling.ts
```diff
@@ -110,7 +111,10 @@ export async function handleWsMessageSubscribeTypes(
   } else if (data.type.includes('ledger')) {
     const current_ledger = data as StreamLedger
     ledger_hashes.push(current_ledger.ledger_hash)
+
     if (networks) {
+      await insertValidatedLedger(networks, current_ledger)
```
Patel-Raj11, line 116: I think we are not de-duplicating when hitting database. Not sure if this will lead to deadlock.
try catch in `insertValidatedLedger` would prevent node.js terminating, but not the deadlock if it ever occurs.

If we want to keep this as it is, I would suggest deploying this branch on staging/dev and then merging it in production if it works well.

ckeshava: image

On line 33, I'm checking against for duplicates against the primary-key of the db. Given that these 3 columns make the primary-key, I expect this operation to be optimized by the DB. 

Besides, I'm not executing any other instruction in that method. Hence, there is no possibility of ungraceful-closure of the db operation (or) leakage of DB-connections in this method.

Sure,  we can observe in dev/stg environment for a few days.

src/connection-manager/connections.ts
```diff
@@ -233,5 +233,23 @@ export default async function startConnections(): Promise<void> {
...
+  }
+}
+
+// delete all validated ledgers older than the 0000 hours mark of the current date.
+async function pruneValidatedLedgersTable(): Promise<void> {
+  const currentDate = new Date()
+  const previousDate = new Date()
+  previousDate.setDate(currentDate.getDate() - 1)
```
Patel-Raj11, line 245: I think you meant to set `-30` since we want to keep past 30 days of entries. 

Something like this:

const cutoffDate = new Date();
cutoffDate.setUTCHours(0, 0, 0, 0) // normalize to UTC midnight
cutoffDate.setUTCDate(cutoffDate.getUTCDate() - 30) // go back 30 days in UTC

src/connection-manager/wsHandling.ts
Patel-Raj11: Not blocking the PR review: I would recommend testing it in staging/dev for a few days.

src/shared/database/validatedLedgers.ts
```diff
@@ -0,0 +1,101 @@
...
+        .insert({
+          ...existingLedgers[0],
+          validation_public_keys: Array.from(validation_public_keys),
+        })
+        .onConflict(['ledger_index', 'ledger_hash', 'network'])
+        .merge({
+          validation_public_keys: Array.from(validation_public_keys),
+        })
+    } else if (existingLedgers.length === 0) {
+      log.error(
```
Patel-Raj11, line 85: Can we print `validation_public_keys` for this case? So that we know which validator sent us the validation message that didn't belong to any ledger from logs?

src/shared/database/setup.ts
Patel-Raj11: Ok, addressing https://github.com/ripple/validator-history-service/pull/370/files#r2323192872 will help us see the validator in logs.

ckeshava: fixed in commit: 94abe5f8cd58f185e972b4ba4f6a5d14732f401f

src/connection-manager/connections.ts
ckeshava: my bad, fixed in commit: 94abe5f8cd58f185e972b4ba4f6a5d14732f401f

src/connection-manager/connections.ts
ckeshava: fixed in commit: 94abe5f8cd58f185e972b4ba4f6a5d14732f401f

src/shared/database/validatedLedgers.ts
ckeshava: fixed in commit: 94abe5f8cd58f185e972b4ba4f6a5d14732f401f

Patel-Raj11 (dismissed):

ckeshava:

PR #373, opened 2025-08-25
persist validations inside the db

ckeshava:

Store the validation messages inside the database. If the VHS-connections service is down, then the persisted validation-messages can be used to accurately calculate the agreement scores. At the present time, the validations are only stored inside the ephemeral RAM memory, which is vulnerable to data loss during pod restarts.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

This PR only stores the validations inside the database. No change has been made to the agreement-score calculation logic yet.

## Test Plan
Appropraite unit tests have been added.

## Future Tasks
1. After our initial investigation, there is an opportunity to purge the validations table more frequently. This will offer significant disk-space-usage reductions in the VHS-connections service.
2. If the pod complains of `KnexPoolTimeout` (or similar errors), then the initial Knex config needs to be updated to a suitable value. Depending on the deployment machine hardware configuration and the needs of the VHS-connections service, we need to set a suitable `max` size for the database connections pool. 

Please [read this document](https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing) before suggesting a very high max-connections number.


ckeshava:

PR #430, opened 2026-01-15
bug: Rectify `validators` table chain misclassification

ckeshava:

The VHS code is mis-classifying the XRPL main-net validators into a different chain. The below picture shows an example of the incorrect classification.

### Root Cause
The [prior work](https://github.com/ripple/validator-history-service/pull/396/changes) failed to account for validators incorrectly configuring their `network_id` values. As an example, the following validator-signing-keys have repeatedly broadcast XRPL Mainnet validations with `network_id === 1 (testnet)`. The VHS was trusting these `network_id` values, hence it causes data corruption.

This PR resolves this issue by taking a majority-vote on the `network_id` proposed for an incoming ledger. This upgrade improves the robustness of the VHS. The newly added test file `test/chains/finalize-ledger-network-id.test.ts` explicitly validates this scenario.

### Reproduction of the Error

Here is an example of the [mis-classification of the Ripple validator](https://vhs.stg.ripplex.io/v1/network/validator/nHU4bLE3EmSqNwfL4AP1UZeTNPrSPPP6FXLKXo2uqfHuvBQxDVKd) in the `chain` key-value. 

Demonstration of the error: The below validator belongs to the XRPL Mainnet, hence `chains` should be `main`.
image

I have deployed this fix-PR into the VHS-dev environment.

### Salient points of this PR
1. This PR solves the issue described above. (`finalizeLedgerValidations` function)
2. It simplifies the logic of classifying ledgers into different XRPL chains. (`saveValidatorChains` function)
3. Introduces Invariant-check mechanisms that Sanity-Checks potential corruptions in Chain-validator data. (`auditChainValidatorsOverlap` function)

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

A new test suite has been added to specifically test the case of incorrect `network_id` configuration amongst a minority of the validators.

Additional verification points:
- The `chains` property of the XRPL Mainnet validators must be `main`. The attached picture shows the incorrect classification of a validator.
- The total number of ledgers processed in the last 1 hour should be ~930 ledgers. On similar lines, the total number of ledgers processed over 24h must be ~22.2k ledgers.
These data points can be viewed via the [VHS APIs](https://vhs.dev.ripplex.io/v1/network/validator/nHBWa56Vr7csoFcCnEPzCCKVvnDQw3L28mATgHYQMGtbEfUjuYyB) (or) via the [appropriate Explorer page](https://livenet.dev.ripplex.io/validators/nHBWa56Vr7csoFcCnEPzCCKVvnDQw3L28mATgHYQMGtbEfUjuYyB/history).

The fix has been deployed since March 5, 2026 on the VHS-dev environment.


ckeshava:
Update: 
The root cause of the issue is that the data inside the `validators` table has been corrupted in the VHS staging environment. This has led to incorrect classification of validators across three XRPL chains (Mainnet, Testnet and xahau-main). It is not clear to me why the data in the `validators` table was corrupted in the first place. I have verified this by gathering the validator-signing-keys across the three XRPl chains. There are several keys which appear in all the three chains -- this is an indication of data corruption.

I have not observed this issue in the VHS-production environment. Hence, this bug must have been introduced in my recent refactor of the network_id related files.

This PR is a step in the right direction. It removes the dependency on reading information from the validators table, for the purposes of assigning validators to different chains. This eliminates the circular dependency of data corruption. We can watch the VHS-dev environment for the next few days as a sanity check.

ckeshava:

PR #439, opened 2026-03-04
fix: assorted amendment fixes

pdp2121:

Make the following fixes to amendment retrieval:
- Update the deprecated column to be true if an amendment in voting returns `badFeature` error in FeatureAll or FeatureOne requests.
-  Add features with supported:true but has not received any vote into the vote API.

### Context of Change

In rippled [v3.1.1](https://github.com/XRPLF/rippled/releases/tag/3.1.1), Batch and fixBatchInnerSigs are unsupported after being voted in 3.0.0. VHS should remove these from voting API.

New amendments on testnet are not enabled by vote, but are still supported, should still be added into the amendment table.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


kuan121 (dismissed):

ckeshava (commented):
Hello Phu, why do we need this change? As you correctly pointed out in the offline discussions, an `unsupported` amendment cannot be voted-on by the validators. Hence, the votes received by an unsupported amendment will remain frozen at the given value.

Why are these changes required?

kuan121:
> @kuan121 @ckeshava can you guys re-review? I have updated the content of this pull request (see updated description) and tested on dev environment.

@pdp2121 What specific manual tests did you perform in the dev environment? Can you share some notes about them?

kuan121 (dismissed):

kuan121 (approved):

ckeshava (approved):
As a future opportunity to simplify the code: I would like to simplify the architecture of the amendment-related code in the VHS. I was hoping to consolidate `amendment_status` and `amendment_info` tables into one single table. I suspect this will also provide simplification in the API layer. Furthermore, I believe we don't need to read the amendment-values from the JSON file because we periodically fetch the AMENDMENTS ledger-object.

PR #501, opened 2026-07-01
Amendment retired/obsolete classification

pdp2121:

- Replace the single `deprecated` flag with two separate fields: `retired` and `obsolete`. Distinction: Retired amendments were enabled long ago and are now permanently baked into the protocol (rippled removed their pre-amendment code), whereas obsolete amendments were never enabled and never will be — they were superseded, pulled, or dropped before ever passing.
- Figure out which is which by reading rippled's `features.macro`:
  - `retired` comes from the latest release tag (so we don't flag amendments before they ship).
  - `obsolete` means the amendment is gone or marked obsolete in both the latest release and develop (so beta amendments aren't wrongly flagged). Also treat `Supported::No` in the latest release as not-votable, a.k.a obsolete (e.g `Batch` in 3.2.0).
- Every amendment in the DB gets reclassified by name, not just the ones the feature RPC still returns, so old removed amendments get flagged too.
- If `features.macro` can't be fetched or parsed, skip the update instead of overwriting flags with wrong values.
- Schema: add `retired`/`obsolete` columns to `amendments_info`, drop `deprecated`. The API now returns both fields.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Verified on dev with new v3.2.0 retirements:
https://livenet.dev.ripplex.io/amendments
https://vhs.dev.ripplex.io/v1/network/amendments/info

### DIFF
--- a/src/shared/database/amendment-classification.ts
+++ b/src/shared/database/amendment-classification.ts
@@ -1,0 +1,145 @@
+import axios from 'axios'
+
+import {
+  EnvironmentVariable,
+  getEnvironmentVariable,
+} from '../utils/environment-variable'
+import logger from '../utils/logger'
+
+const log = logger({ name: 'amendment-classification' })
+
+// The classification of an amendment as "retired" or "obsolete" is only
+// authoritatively declared in rippled's features.macro file:
+//   - XRPL_RETIRE_FEATURE / XRPL_RETIRE_FIX  -> retired (active 2+ years, code
+//     removed, identifier deprecated, permanently baked in).
+//   - VoteBehavior::Obsolete                 -> obsolete (supported but never
+//     passed, kept votable but marked so no one votes for it).
+// Neither is reliably derivable from the (public) feature RPC, so we parse the
+// file directly. We pin to a stable rippled release tag - NOT `develop` - so
+// the classification only changes when we intentionally bump the tag.
+const DEFAULT_RIPPLED_RELEASE_TAG = '3.2.0'
+
+/**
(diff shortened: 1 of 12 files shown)


src/shared/database/amendment-classification.ts
```diff
@@ -0,0 +1,267 @@
...
+ * @param url - The raw features.macro URL.
+ * @returns The parsed features.macro, or null on failure.
+ */
+async function fetchFeaturesMacro(
+  url: string,
+): Promise<ParsedFeaturesMacro | null> {
+  try {
+    const response = await axios.get<string>(url, { responseType: 'text' })
+    const parsed = parseFeaturesMacro(response.data)
+    if (parsed.retired.size === 0 && parsed.all.size === 0) {
```
cybele-ripple: In the current form, the only time this would be true is if `parsed.all.size === 0` (since `parsed.all.size === 0`  implies `parsed.retired.size===0`). Do you mean to use `if (parsed.all.size === 0 || parsed.retired.size === 0` instead? Using `||` would allow cases when `parsed.retired` is empty but not `parsed.all`

pdp2121: Great catch! Fixed in [fbd0b0e](https://github.com/ripple/validator-history-service/pull/501/commits/fbd0b0e5b7a0eb858b270efda12fe0b4d5a9561d)

kuan121 (approved):

cybele-ripple (approved):

PR #509, opened 2026-09-09
fix: validator reports for validators without a master key

pdp2121:

For `/reports` endpoint for validator:

- Join on `COALESCE(validators.master_key, validators.signing_key)`
- Match `master_key` or `signing_key`, mirroring `findInDatabase` in `validator.ts`

### Context of Change

`GET /v1/network/validator/:publicKey/reports` returned `count: 0` for validators that never published a manifest, even though their agreement data exists.

Agreement rows are **written** keyed by `master_key ?? signing_key` (`saveDailyAgreement`, `src/connection-manager/agreement.ts`), but were **read** keyed on `master_key` alone:

```ts
.innerJoin('validators', 'daily_agreement.main_key', 'validators.master_key')
.where('validators.master_key', '=', master_key)
```

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

- `reports` shows up for validator without manifest on dev (https://vhs.dev.ripplex.io/v1/network/validator/n9LS8sE7BCxYFJKC2eR3dFLGLRu7KYq2s88z1irdUTUkha4qWv4q/reports)

### DIFF
--- a/src/api/routes/v1/validator-report.ts
+++ b/src/api/routes/v1/validator-report.ts
@@ -1,10 +1,15 @@
 import { Request, Response } from 'express'
 
-import { query } from '../../../shared/database'
+import { db, query } from '../../../shared/database'
 import { AgreementScore } from '../../../shared/types'
 import logger from '../../../shared/utils/logger'
 
 const log = logger({ name: 'api-validator-report' })
+
+// Agreement rows are keyed by `master_key ?? signing_key` (see
+// `saveDailyAgreement` in the connection-manager), so reads must resolve a
+// validator to that same effective key.
+const EFFECTIVE_KEY = 'COALESCE(validators.master_key, validators.signing_key)'
 
 interface ScoreResponse {
   validation_public_key: string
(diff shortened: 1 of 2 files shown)


src/api/routes/v1/validator-report.ts
```diff
@@ -1,11 +1,16 @@
...
-import { query } from '../../../shared/database'
+import { db, query } from '../../../shared/database'
 import { AgreementScore } from '../../../shared/types'
 import logger from '../../../shared/utils/logger'
 
 const log = logger({ name: 'api-validator-report' })
 
+// Agreement rows are keyed by `master_key ?? signing_key` (see
+// `saveDailyAgreement` in the connection-manager), so reads must resolve a
+// validator to that same effective key.
```
kuan121, line 11: > Pre-manifest history gap (inherent to the main_key design, not a regression). If a validator produced agreement rows before publishing a manifest (keyed by signing_key) and later published one (master_key set), COALESCE now resolves to master_key, so those older signing-key-keyed rows won't join. The old query had this same limitation (worse, actually — it matched nothing without a master key). The PR neither introduces nor is expected to fix this; just be aware historical continuity across a first-manifest event isn't guaranteed on either side.

This is a review comment from Claude Code, and I think it makes sense. Should we handle it in this PR or a follow-up one?

src/api/routes/v1/validator-report.ts
```diff
@@ -54,23 +59,29 @@ function formatResponse(response: DatabaseResponse): ScoreResponse {
...
- * @param master_key - Master key of validator.
+ * Accepts either a master key or a signing key, matching the behavior of the
+ * validator endpoint. Validators that never published a manifest have a null
+ * `master_key` and are only reachable by their signing key.
+ *
+ * @param public_key - Master key or signing key of validator.
  * @returns A promise that resolves to an array of ScoreResponse.
  */
-async function getReports(master_key: string): Promise<ScoreResponse[]> {
+async function getReports(public_key: string): Promise<ScoreResponse[]> {
```
kuan121, line 69: > Sibling endpoint has the identical latent bug (out of scope). [daily-report.ts:82-88](vscode-webview://1lsqbmokbbe2ojg5mpedsg779mbg6bt2114pbkpcjsgicgdtmjo3/src/api/routes/v1/daily-report.ts#L82-L88) (the /daily_report endpoint) joins main_key = master_key and adds .whereNotNull('validators.master_key'), so it silently omits every no-manifest validator too. Not this PR's job to fix, but the same class of bug lives there — worth a follow-up ticket so the two endpoints don't drift.

This is a review comment from Claude Code, and I think it makes sense. Should we handle it in this PR or a follow-up one?

kuan121 (approved):
Approving for now since this fixes the immediate issue, but it doesn’t cover some edge cases mentioned here: https://github.com/ripple/validator-history-service/pull/509#discussion_r3974366691, as well as a related issue mentioned here: https://github.com/ripple/validator-history-service/pull/509#discussion_r3974369433.

src/api/routes/v1/validator-report.ts
pdp2121: This is valid for consistency. Fixed in [2f1bfc7](https://github.com/ripple/validator-history-service/pull/509/commits/2f1bfc7cba01ed0300000315f5c32c147d6dff14)

src/api/routes/v1/validator-report.ts
pdp2121: This is out-of-scope for this PR. I will leave it as follow up

sguntaka1578-ripple (approved):

kuan121 (approved):

